Issue #1, opened 2024-10-01
RFC: Plan for establishment of xsf library for special function scalar kernels


steppi:
### Some background and history

In August of 2023 Irwin Zaid (@izaid) made the following [comment](https://github.com/scipy/scipy/pull/19023#issuecomment-1689050946) on the pull request https://github.com/scipy/scipy/pull/19023,

> Something tangential: I know this issue is about the array API, but it might be good to explore whether the C versions of the special functions can be made very easily "copy-and-pasteable" to include in other array libraries. This way, although the C implementation isn't a defined standard itself, it would help out a lot of other projects if SciPy has something other libraries can easily steal. Example: how much of CuPy's special implementation is just taken directly from SciPy's, with `__device__` in front of everything? Could SciPy just have a lot of header files for special functions that other projects can easily include?

For context: the array API, short for the [Python array API standard](https://data-apis.org/array-api/latest/), was proposed to reduce fragmentation in the Python array computing world by offering a common API for an important subset of array computing functionality. The goal is that code written to the array API standard should work with any conforming array backend (e.g. NumPy, PyTorch, CuPy, Jax, etc.). Although the array API standard is restricted to a core set of important features, there can be optional[extensions](https://data-apis.org/array-api/latest/extensions/index.html), standardizing *coherent sets of functionality*. There are currently extensions for [Linear Algebra](https://data-apis.org/array-api/latest/extensions/linear_algebra_functions.html) and [Fourier transform functions](https://data-apis.org/array-api/latest/extensions/fourier_transform_functions.html). It seems natural that there should be a special function extension[^1] codifying consistent naming and argument conventions for a set of the most widely used special functions, but due to inconsistent support for special functions across array libraries, an extension restricted only to functions which are currently uniformly available among the most widely used array libraries would be missing many useful functions. It is much easier for these array libraries to add a new special function when there is an existing implementation that can be re-used with at most small modifications.

At the time, SciPy could not provide header files for many of its special functions that other projects can easily include because SciPy special was a polyglot submodule with special function scalar kernels written in each of C, C++, Cython, and Fortran 77. Scalar kernels written in Cython and Fortran 77 would require translation to run on the C/C++ based CUDA platform for use in GPU aware array libraries, and we observed that it was precisely the functions with such scalar kernels that were most likely to be missing from other array libraries.

[...]

PR #4, opened 2024-11-13
TST: Add CMake and gtest frameworks

inkydragon:
This is a POC that currently sets up a basic CMake + gtest test environment.

- [x] CMake
- [x] gtest
- [x] github CI, Generate and upload coverage report

Build and generating coverage report:
```sh
# sudo apt-get install -y lcov

cmake -DCMAKE_BUILD_TYPE=Coverage -S . -B build
cd build && make && ctest
make coverage_html

# build\coverage_report
```

Sample:
- [Codecov report](https://app.codecov.io/gh/inkydragon/xsf/tree/main/include%2Fxsf)
	- Need to set up codecov CI and patch: [inkydragon/xsf@`73c842f` (#1)](https://github.com/inkydragon/xsf/pull/1/commits/73c842f8c9bcedd8d63aedda8eb0f8d321fd9b6d)
	- Only a small number of files are shown in the report 
	because the test code only calls the airy-related functions
- [HTML coverage report](https://github.com/inkydragon/xsf/actions/runs/11814616780)

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -3,7 +3,8 @@
 include(FetchContent)
 FetchContent_Declare(
   googletest
-  URL https://github.com/google/googletest/releases/download/v1.15.2/googletest-1.15.2.tar.gz
+  GIT_REPOSITORY https://github.com/google/googletest.git
+  GIT_TAG        b514bdc898e2951020cbdca1304b75f5950d1f59  # v1.15.2
 )
 # For Windows: Prevent overriding the parent project's compiler/linker settings
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -11,6 +11,10 @@ FetchContent_MakeAvailable(googletest)
 include(GoogleTest)
 
 # Add Tests
-add_executable(airy airy.cpp)
-target_link_libraries(airy GTest::gtest_main)
-gtest_discover_tests(airy)
+function(add_gtest test_name)
+    add_executable(${test_name} ${test_name}.cpp)
+    target_link_libraries(${test_name} GTest::gtest_main)
+    gtest_discover_tests(${test_name})
+endfunction()
+
+add_gtest(airy)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -8,7 +8,9 @@ FetchContent_Declare(
 # For Windows: Prevent overriding the parent project's compiler/linker settings
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)
 FetchContent_MakeAvailable(googletest)
+
 include(GoogleTest)
+include(${CMAKE_SOURCE_DIR}/tests/Coverage.cmake)
 
 # Add Tests
 function(add_gtest test_name)
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,0 +1,16 @@
+# Get googletest
+# https://google.github.io/googletest/quickstart-cmake.html
+include(FetchContent)
+FetchContent_Declare(
+  googletest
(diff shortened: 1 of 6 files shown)


steppi:
Thanks @inkydragon. This aligns well with what I had planned. My thoughts for testing is that we should have scripts that generate csv files of test cases. The C++ tests would then read these files, and we would have different tiers of tests. Some running on every push to a PR, and others (like GPU jobs or more extensive tests) running on a cron.

For the short term I plan to stash the csv files in GitHub, but eventually they would likely go somewhere better suited for file storage. I'll start generating some files this week, and if you want to continue this, you can either write something to pull them in for the tests, or I can follow through on what you're setting up to do that.

tests/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
+# Get googletest
+# https://google.github.io/googletest/quickstart-cmake.html
+include(FetchContent)
+FetchContent_Declare(
+  googletest
+  URL https://github.com/google/googletest/releases/download/v1.15.2/googletest-1.15.2.tar.gz
+)
```
lucascolley, line 7: Locally I see

```
CMake Warning (dev) at .pixi/envs/dev/share/cmake-3.30/Modules/FetchContent.cmake:1373 (message):
  The DOWNLOAD_EXTRACT_TIMESTAMP option was not given and policy CMP0135 is
  not set.  The policy's OLD behavior will be used.  When using a URL
  download, the timestamps of extracted files should preferably be that of
  the time of extraction, otherwise code that depends on the extracted
  contents might not be rebuilt if the URL changes.  The OLD behavior
  preserves the timestamps from the archive instead, but this is usually not
  what you want.  Update your project to the NEW behavior or specify the
  DOWNLOAD_EXTRACT_TIMESTAMP option with a value of true to avoid this
  robustness issue.
Call Stack (most recent call first):
  tests/CMakeLists.txt:4 (FetchContent_Declare)
This warning is for project developers.  Use -Wno-dev to suppress it.
```

inkydragon:
> I can follow through on what you're setting up to do that.

Feel free to open a new pr based on this.
Or you can open a new pr and give me an example: generate CSV, read and test a function
Then I can follow the steps in that example and get on with the boring test generation.

I guess there are a lot of pr in scipy waiting for your review.

My previous plan was to manually write some test cases to achieve higher line of code coverage with a small number of tests.

tests/CMakeLists.txt
inkydragon: Switch to use the commit id to eliminate this warning

ev-br:
I'm sure y'all thought about this already, but I'll ask nonetheless: have you considered adding a thin python layer for testing and using python testing frameworks? 

While I don't have a horse in this race, the setup otherwise seems suspiciously of the "generate text data and use gnuplot" type of construction which got ditched once somebody showed me `matpltlotlib` back in the day.

steppi:
> I'm sure y'all thought about this already, but I'll ask nonetheless: have you considered adding a thin python layer for testing and using python testing frameworks?
> 
> While I don't have a horse in this race, the setup otherwise seems suspiciously of the "generate text data and use gnuplot" type of construction which got ditched once somebody showed me `matpltlotlib` back in the day.

Yeah, using Python is the other option I've considered. I  suggested that in #1 because I thought it would be the easiest way to get started. In any case, I'd want us to generate test cases and store them somewhere because generating arbitrary precision reference data takes several orders of magnitude longer than running the C++ code. In terms of how to serialize the test cases, I like text formats, because then it's clear they are what they say they are. Binary files could be compromised potentially.

If the test cases are coming from files, and the C++ tests just read files, call the function on each test case, and compare to the reference value, I think that's pretty straightforward. Having a C++ testing framework also makes it easier to test internal components which aren't so easy to wrap in Python like ufuncs are.

I'm really happy to see @inkydragon's interest in helping us get set up with CMake and gtest.

lucascolley:
FYI @inkydragon we've now merged in CMake support, including coverage. Feel free to suggest any improvements!

PR #5, opened 2024-11-13
STY: clang-format repo

lucascolley:
Continues https://github.com/scipy/scipy/pull/21640.

cc @fancidev


lucascolley:
@steppi I know there are more important things for you to prioritise, but it would be great if you could check whether any of the formatting configuration looks unacceptably bad here. I think that even getting a non-ideal configuration merged is a big step up from having none at all, as that will allow us to enforce it in CI and make all future formatting automatic.

lucascolley:
the diff is scarily large just due to `amos.h`, `specfun.h`, and `faddeeva.h` using different indentation to the rest of the codebase

.clang-format
```diff
@@ -6,7 +6,9 @@ BreakBeforeBraces: Attach
 Cpp11BracedListStyle: true
 NamespaceIndentation: Inner
 AlwaysBreakTemplateDeclarations: true
-SpaceAfterCStyleCast: true
+SpaceAfterCStyleCast: false
 ColumnLimit: 120
 InsertNewlineAtEOF: true
 AlignAfterOpenBracket: BlockIndent
+IncludeBlocks: Preserve
+AllowShortFunctionsOnASingleLine: None
```
steppi, line 14: I seem to remember that @izaid deliberately chose a style where short functions could be on a single line. Is there any reason we are changing this now. It mostly effects wrappers like those appearing in config.h.

lucascolley: If I remember correctly, this came from discussions with @fancidev on the superseded PR to SciPy. It was a while ago, though, so not sure

include/xsf/amos/amos.h
```diff
@@ -94,6152 +94,6501 @@
...
-        /* 20 */
-        if (az >= rl) {
-            //
-            // ASYMPTOTIC EXPANSION FOR LARGE Z FOR THE I FUNCTION
-            //
-            nw = asyi(zn, fnu, kode, nn, y, rl, tol, elim, alim);
+    int acai(std::complex<double>, double, int, int, int, std::complex<double> *, double, double, double, double);
+    int
+    acon(std::complex<double>, double, int, int, int, std::complex<double> *, double, double, double, double, double);
+    int asyi(std::complex<double>, double, int, int, std::complex<double> *, double, double, double, double);
```
steppi, line 107: It seems a little noisy that clang-format will sometimes put the return type on the previous line and sometimes on the same line, in order to keep from exceeding the max column width. I think we should just pick one style and be consistent. Can you see how it looks after setting `BreakAfterReturnType` in .clang-format  to `None`?

.clang-format
steppi, line 1: Standard should be `Cpp17`.

include/xsf/amos/amos.h
lucascolley: done

.clang-format
lucascolley: you can take a look at what it looks like without adding this option now

lucascolley:
if you would like to test other config options locally, you can do so with `pixi run format`

.clang-format
steppi: I see. From here, https://github.com/scipy/scipy/pull/21640#discussion_r1782534326, it looks like @fancidev only wanted to avoid having short functions automatically converted to one line, but didn't want existing one line wrappers to become split across multiple lines. That's good, it seems everyone is actually in agreement about not including `AllowShortFunctionsOnASingleLine: None`. I didn't want to change something that @izaid had explicitly chosen during a period of time where he doesn't have time to participate in discussions.

include/xsf/amos/amos.h
steppi: Huh, that didn't work though. There's still a mix.

steppi: I see, `None` is deprecated, and now it does something called `Automatic` that sometimes does and sometimes doesn't break depending on a penalty calculation. I don't want to fight clang-format so let's just take out `BreakAfterReturnType`.

.clang-format
```diff
@@ -1,12 +1,15 @@
...
 NamespaceIndentation: Inner
 AlwaysBreakTemplateDeclarations: true
-SpaceAfterCStyleCast: true
+SpaceAfterCStyleCast: false
 ColumnLimit: 120
 InsertNewlineAtEOF: true
 AlignAfterOpenBracket: BlockIndent
+IncludeBlocks: Preserve
+BreakAfterReturnType: None
+# AllowShortFunctionsOnASingleLine: None
```
steppi, line 15:
```suggestion
```

steppi:
I made a final update to the clang-format config, cleaned up history with an interactive rebase, and added a commit to add the clang-format commit to `.git-blame-ignore-revs` https://github.com/scipy/xsf/compare/main...steppi:xsf:format.
If everything is in order, I plan to force push to this branch and merge this.

steppi (approved):
I think it's good enough. Let's ship it.

lucascolley:
thanks Albert! Now lets get the submodule PR in before cherry-picking becomes a nightmare 😅

PR #14, opened 2025-03-11
TST: Add data driven tests based on `xsref` parquet files using Catch2

steppi:
This PR is a rough sketch of what I'm thinking of for tests. It uses Catch2 as the testing framework, and implements a custom Catch2 generator to iterate through test cases. Tests are only implemented for `hyp2f1` on this PR. Once I have everything settled down, I plan to write a script to generate the test files for other functions since they follow a pretty straightforward recipe. I've only added tests for 64 bit Linux GCC, but more can come later. I'll need to add a parquet file of tolerances for every tested platform, but this shouldn't be too bad actually.

I think Catch2 seems nice enough to work with. So far I've found one thing that it would be good to have fixed. It seems that all floating point numbers are printed with `std::fixed` when using Catch2's built in logging features, when `std::scientific` is needed for our case in order to actual see floating point values to full precision. There's a languishing PR, https://github.com/catchorg/Catch2/pull/2226, to allow setting the format to `std::scientific`, which only needs a small tweak to make work. I see this as a bellweather. If I'm able to nudge this PR through on a reasonable timescale, then I think Catch2 is actively developed enough to be worth using here, but if the PR continues to languish, I guess we can just use GoogleTest or something. I'll leave a comment with more details on this.

I haven't done any of the nice stuff with coverage that @inkydragon did. This is important, but this is just a sketch for now. @inkydragon, let me know if you think there's anything else that could be improved here. It also might be a good idea to set up ccache.

 I'm not sure if github workflow i wrote will work yet. I still haven't got up to speed on testing those locally. The tests themselves due work locally for me though.


tests/test_hyp2f1.cpp
```diff
@@ -0,0 +1,67 @@
...
+    tol = adjust_tolerance(tol);
+    INFO("a := " << std::setprecision(std::numeric_limits<double>::max_digits10) << a << '\n'
+	 << "b := " << b << '\n'
+	 << "c := " << c << '\n'
+	 << "z := " << z << '\n'
+	 << "out := " << out << '\n'
+	 << "desired := " << desired << '\n'
+	 << "error := " << error << '\n'
+	 << "tolerance := " << tol << '\n'
+	);
```
steppi, line 40: If https://github.com/catchorg/Catch2/pull/2226 lands. This could be replaced with

```c++
CAPTURE(a, b, c, z, out, desired, error, tolerance);
```

steppi:
OK. Well we have tests for one platform now. The parquet files contain cases from all `scipy.special` tests which I generated using a series of scripts from this `xsref` PR https://github.com/scipy/xsref/pull/1. I used another script from that `xsref` PR to generate the C++ test files. 

I had to make some fixes to get all of the tests to pass, including the problems noted by @ZappD0S  here https://github.com/scipy/xsf/issues/16. So this PR should close that issue.

Some remaining work for testing and CI:

1. Add documentation for how to run the tests.
2. Add test coverage like what was done in https://github.com/scipy/xsf/pull/7
3. Use `ccache` in CI to save time.
4. Add error tolerance parquet files for more platforms.
5.  Test more platforms on CI.
6. Workflows to update platform specific error tolerance parquet files for continuous improvement.

I plan to work on these things incrementally, spending a few hours per week improving the state of things.

steppi:
I'm going to go ahead and merge this. I think we can iterate from here.

PR #19, opened 2025-04-05
CI: test macOS

lucascolley:
TODO:
- fix Linux errors:
```
	 32 - cyl_bessel_i dd->d scipy_special_tests (Failed)
	 33 - cyl_bessel_i dD->D scipy_special_tests (Failed)
```
- set up tolerances for macos


.gitignore
```diff
@@ -43,3 +43,6 @@ build/
 
 # xsref
 xsref/
+
+# IDEs
+.vscode
```
lucascolley, line 48:
```suggestion
.vscode

```

pixi.toml
```diff
@@ -63,10 +62,23 @@ configure-tests.env.XSREF_TABLES_PATH = "$PWD/xsref/tables"
...
+
+[feature.tests-ci.tasks]
+build-tests-ci.cmd = ["cmake", "--build", "build", "-j3"]
+build-tests-ci.depends-on = ["clone-xsref", "configure-tests"]
+build-tests-ci.cwd = "."
+
+# run tests
+tests-ci.cmd = ["ctest", "--output-on-failure", "--test-dir", "build/tests", "-j3"]
+tests-ci.depends-on = ["build-tests-ci"]
+tests-ci.cwd = "."
```
lucascolley, line 80: I've hardcoded `-j3` just for CI here so that we can keep the CI invocation a one-liner. The parallelism is now controlled in the manifest rather than the workflow file, which seems fine to me, but happy either way. You can always run the individual tasks with different `-j` values locally in the CI env to reproduce failures more quickly like e.g.

```shell
pixi run --environment=tests-ci build-only -j24
```

tests/testing_utils.h
```diff
@@ -70,6 +70,10 @@ class TableReader {
             V imag;
             *stream_ >> real >> imag;
             element = U(real, imag);
+        } else if constexpr (std::is_same_v<U, long>) {
+            int64_t temp;
+            *stream_ >> temp;
+            element = static_cast<U>(temp);
```
lucascolley, line 76: disclaimer: LLM generated, but I understand that it is checking for `long`'s and casting via `int64_t` first to be compatible with `StreamReader`

lucascolley (commented):
@steppi here are all the changes needed to get macOS tests running in CI. Of course, the tolerances need adjustments.

---

As a separate bit of fun, CI is now exactly reproducing the errors I saw on my linux box.

tests/testing_utils.h
steppi: Yeah. That's it. I think LLM code won't work on all platforms, since `long` won't always be `64 bit`. I made essentially the same suggestion here https://github.com/scipy/xsf/pull/6#issuecomment-2780758436, but using `std::int64_t`.

pixi.toml
lucascolley: I don't think so, but can this sort of parallelism be controlled via environment variables instead? If so, that is most convenient (can set a default in the Pixi tasks which is overridable in CI).

lucascolley: actually, I think we can do this, let me give it a go

lucascolley: looks like we want https://github.com/prefix-dev/pixi/pull/3433, which was just merged 2 days ago :) let's not block this PR then, but open an issue to use arguments for parallelism once that PR is released

lucascolley:
The default compiler for macOS from conda-forge looks to be `Clang 18.1.8`

.gitignore
steppi, line 48: Do you not set up a global `.gitignore` with your editor specific stuff? https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

I'd rather encourage devs to do that rather than try to add every single IDE specific thing here.

lucascolley: Oh sure, I can do that if you prefer. I guess I've generally been happy with people putting anything in `.gitignore`

lucascolley:
okay, I see

```
75% tests passed, 58 tests failed out of 231
```

locally, let's see if CI agrees

tests/testing_utils.h
```diff
@@ -127,25 +131,20 @@ Catch::Generators::GeneratorWrapper<std::tuple<T1, T2, T3>> xsf_test_cases(
...
     // Add some wiggle room to tolerance from table.
     return 4 * std::max(std::numeric_limits<T>::epsilon(), tol);
 }
 
-
 std::string get_platform_str() {
     /* This is going to get a string "<compiler>-<os>-<architecture>" using conditional
      * compilation, but for now we're just stubbing things out. */
     return "gcc-linux-x86_64";
 }
```
steppi, line 144: I added the tolerance files for Mac, but we need to use preprocessor directives here to make sure the correct file gets selected for each platform. I was looking at this, https://stackoverflow.com/questions/152016/detecting-cpu-architecture-compile-time.

tests/testing_utils.h
```diff
@@ -127,25 +131,20 @@ Catch::Generators::GeneratorWrapper<std::tuple<T1, T2, T3>> xsf_test_cases(
     );
 }
 
-
 template <typename T>
 T adjust_tolerance(T tol) {
     // Add some wiggle room to tolerance from table.
     return 4 * std::max(std::numeric_limits<T>::epsilon(), tol);
```
steppi, line 137: For failures due to different compiler versions, like the currently failing Linux tests, I think it's fine just to increase the wiggle room factor here for now, as long as it doesn't get too large.

tests/testing_utils.h
lucascolley: Okay, cool.

Long-term, we could look at specifying the compiler version here as well? Thinking ahead, having a shared Pixi workspace between xsf and xsfref should allow the reference values to be generated by the same envs as they are tested in against the implementations.

steppi: Yeah. I originally did specify the compiler version but was trying to avoid having too many files. Long term I think we should be using some kind of cloud storage solution instead of stashing the parquet files on Github, and then it should be fine to get very granular with the platforms/compilers/whatever in the tolerance files. For now, maybe it's fine to just allow individual adjustments of the wiggle factor in `adjust_tolerance` in tests.

steppi:
The 3 Mac OS failures look real.

steppi:
I just yanked those failing tests out. `bdtr`, `bdtri`, and `bdtrc` for non-integer `n` don't even exist in `xsf`, those overloads are defined in SciPy (and are deprecated there). There are other cases like this I think, but I'm just doing the minimum to get tests to pass. The corresponding tables should be removed from `xsref` too.

steppi (approved):
CI is passing now. Thanks @lucascolley! I'm going to go ahead and merge this so we can continue iterating from here.

PR #23, opened 2025-04-05
CI: set up ccache

lucascolley:
on top of gh-19

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -86,6 +86,8 @@ configure-coverage.cmd = [
 ]
 configure-coverage.cwd = "."
 configure-coverage.env.XSREF_TABLES_PATH = "$PWD/xsref/tables"
+configure-coverage.env.CC = "ccache $CC"
+configure-coverage.env.CXX = "ccache $CXX"
 # Open coverage report
 open-coverage.cmd = ["open", "index.html"]
 open-coverage.cwd = "build/coverage_report"
--- a/pixi.toml
+++ b/pixi.toml
@@ -101,7 +101,6 @@ ccache = ">=4.11.2,<5"
 # Build and generate coverage report
 # TODO: use a task arg for parallelism https://github.com/prefix-dev/pixi/pull/3433
 build-tests-ci.cmd = ["cmake", "--build", "build", "-j3"]
-build-tests-ci.env = { CC = "ccache $CC", CXX = "ccache $CXX" }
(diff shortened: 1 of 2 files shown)


lucascolley:
@steppi I tried following the config in SciPy, but it looks like ccache isn't having an impact here

steppi:
> @steppi I tried following the config in SciPy, but it looks like ccache isn't having an impact here

Thanks for checking. CI time is still pretty short anyway so let's leave investigating this for the future.

rgommers:
@lucascolley conda-forge compilers set `CFLAGS` et al on activation, which overwrites what you put into a GHA yml file. What you seem to be missing here is `env = {CC: cache $CC}` in `pixi.toml. See how `gpu.ci.yml` and the corresponding `pixi.toml` in the main scipy repo are set up.

lucascolley:
thanks Ralf, that did the trick!

> Successful in 1m

1m58s, but we'll take it :) 🚀🚀

The bottleneck now seems to be processing the `.gcda` files for coverage, not sure if there is a way for cache to help us there.

pixi.toml
```diff
@@ -99,6 +101,7 @@ ccache = ">=4.11.2,<5"
 # Build and generate coverage report
 # TODO: use a task arg for parallelism https://github.com/prefix-dev/pixi/pull/3433
 build-tests-ci.cmd = ["cmake", "--build", "build", "-j3"]
+build-tests-ci.env = { CC = "ccache $CC", CXX = "ccache $CXX" }
```
lucascolley, line 104: this line may be unnecessary, not sure whether the env vars only need to be set for the configure step

steppi: Yeah, I think you're right. The environment variables should get baked in during the configure step.

lucascolley:
```suggestion
```

steppi:
The failures are because there were some tests for overloads which don't actually exist in `xsf` and I removed the parquet files for them. The tests themselves are removed in #21. Let's just merge that one first, I'll have the tolerance files ready shortly.

steppi:
I merged #21. After merging main this should be ready.

lucascolley:
yep, this is ready. We should see whether `-j2` is any better than `-j3` at some point, but that can probably wait until https://github.com/prefix-dev/pixi/issues/3538 is fixed, 2 minutes is already pretty speedy

PR #25, opened 2025-04-17
Pin xsref version in clone-xsref command

steppi:
This PR pins the version of `xref` in the `clone-xsref` command to `v0.0.0`, which I just created it seems like a good idea to pin to a tag to make it easier to keep things in sync. I can imagine a situation where we add a new batch of tests, but don't want to pin to the newest `xsref` version on `main` until we have them all passing. Also, this will keep tests for old versions failing if parquet files get removed, renamed, or restructured.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -44,7 +44,7 @@ libarrow-all = ">=19.0.1,<20"
 # clean xsref dir
 clean-xsref = { cwd = ".", cmd = "rm -rf xsref" }
 # clone xsref
-clone-xsref.cmd = "git clone --branch v0.0.0 https://github.com/scipy/xsref.git"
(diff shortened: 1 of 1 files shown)


steppi:
> For slightly more flexibility, we could split it up into separate clone and checkout steps, so one can change the version without having to reclone. But fine to just merge this!

Yeah, I was considering that too. 

> For slightly more flexibility, we could split it up into separate clone and checkout steps, so one can change the version without having to reclone. But fine to just merge this!

That's good thinking, but I want to go in a different direction. I just made the clone shallow to save time:

Clone time on my machine went from 

```
real    0m37.408s
user    0m13.567s
sys     0m7.379s
```

to

```
real    0m11.133s
user    0m3.803s
sys     0m2.423s
```

and things would only get worse as `xsref` gets more history.

lucascolley:
Yep, that's a good argument!

PR #28, opened 2025-04-17
add meson xsf_dep

lucascolley:
see https://github.com/steppi/scipy/pull/2

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -1,3 +1,10 @@
-project('xsf')
+project(
+  'xsf',
+  'cpp',
+  version : '0.1.0',
+  license : 'BSD-3-Clause',
+  license_files : 'LICENSE',
+  meson_version : '>=1.5.0'
+)
 
-xsf_dep = declare_dependency(include_directories: 'include')
+xsf_dep = declare_dependency(include_directories : 'include')
--- a/meson.build
+++ b/meson.build
@@ -1,0 +1,3 @@
+project('xsf')
+
+xsf_dep = declare_dependency(include_directories: 'include')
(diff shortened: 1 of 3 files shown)


rgommers (commented):
I think the `project` call needs to use `version`, `license` (as an SPDX expression please), `license_files` and `meson_version` - see https://mesonbuild.com/Reference-manual_functions.html#project for docs.

Other than that, agreed that this is a good idea.

steppi:
> let us know how you want to approach versioning @steppi

I'm not sure. I was thinking the first version of `xsf` that gets included as a submodule in SciPy would be `v0.0.0`. I figured we could create the tag when `xsf` was otherwise in a state where we're ready to merge the submodule (now subproject) PR into SciPy.

rgommers:
I'd recommend at least `0.1.0`, code is pretty mature and well-tested already.

steppi:
> I'd recommend at least `0.1.0`, code is pretty mature and well-tested already.

That's true. The code is mature and well tested on CPU, so I'm fine if you think that's worth `v0.1.0`. My thoughts were

1. There's currently no documentation describing what functions are provided and what headers to include to use them.
2. The headers aren't well organized. They mostly reflect the historical structure of `scipy.special`, which makes the library hard to navigate and use independently.

My thinking was `v0.0.0` would be a signal that while the code is ready to use in SciPy, it's not really ready for broader consumption yet.

lucascolley:
> as an SPDX expression please

FYI, SciPy's `BSD-3` (https://github.com/scipy/scipy/blob/052981390aeacc94a45c620e00a4b1e2224c0606/meson.build#L5) doesn't appear on the SPDX license list (https://spdx.org/licenses/).

`BSD-3-Clause` does, though, so I'll use that here

lucascolley:
Ready for merge!

steppi (approved):
Thanks @lucascolley for doing this and giving a great explanation for why we should do it. Looks good to me.

rgommers:
`BSD-3-Clause` is indeed correct, but more is needed - I filed gh-29.

Issue #29, opened 2025-04-18
Licensing issues in this repository


rgommers:
The current `LICENSE` file is a copy of the SciPy license only, however there are multiple other licenses needed to cover this code - see the entries for `scipy/special` components in https://github.com/scipy/scipy/blob/052981390aeacc94a45c620e00a4b1e2224c0606/LICENSES_bundled.txt

To fix this, I suggest starting by creating a LICENSES_bundled in this repo, and also updating the SPDX expression in the `meson.build` file that was just added - it should be `BSD-3-Clause AND ... AND ...` with the correct license identifiers for the bundled code.

The file paths in SciPy's `LICENSES_bundled.txt` were also never updated for the path changes due to the introduction of `xsf` it looks like. That can be addressed separately or in https://github.com/scipy/scipy/pull/22854

PR #30, opened 2025-04-18
fix licensing issues

lucascolley:
closes gh-29


LICENSES_bundled.txt
```diff
@@ -0,0 +1,119 @@
+xsf bundles several libraries that are compatibly licensed. We list these here.
+
+Name: Cephes
+Files: inlcude/xsf/cephes/*
```
steppi, line 4:
```suggestion
Files: include/xsf/cephes/*
```

lucascolley:
> I guess the reason we don't need put anything here for amos and specfun is because they are public domain?

Not sure, but I don't think we have licenses for them bundled in SciPy

LICENSES_bundled.txt
```diff
@@ -0,0 +1,119 @@
...
+  written license agreement for such Enhancements, then you hereby grant
+  the following license: a non-exclusive, royalty-free perpetual license
+  to install, use, modify, prepare derivative works, incorporate into
+  other computer software, distribute, and sublicense such enhancements
+  or derivative works thereof, in binary and source code form.
+
+Name: mdspan
+Files: include/xsf/third_party/kokkos/mdspan.hpp
+License: Apache-2.0 WITH LLVM-exception
+  For details, see scipy/special/special/third_party/kokkos/mdspan.hpp
```
steppi, line 119:
```suggestion
  For details, see include/xsf/third_party/kokkos/mdspan.hpp
```

rgommers (approved):
That all looks correct to me, thanks for the quick fix!

lucascolley:
thanks for catching the typos Albert :)

steppi (approved):
Thanks @lucascolley!

Issue #37, opened 2025-05-14
Transfer test compiles in CuPy from SciPy


j-bowhay:
xref https://github.com/scipy/scipy/pull/22984 and https://github.com/scipy/scipy/issues/22942

PR #39, opened 2025-06-08
ENH:TST Make changes needed to make xsf work as CuPy submodule and add proof of concept CuPy tests.

steppi:
This PR makes a couple minor changes to `xsf/include/xsf/config.h` needed to make `xsf` work as a submodule of CuPy. 

* `invoke_result` does not work in CuPy because CuPy still compiles its ufunc kernels with language standard C++11. When I make a PR to CuPy to add xsf as a submodule, I will also ask about increasing this, either by adding an ability to pass compiler flags in `cupy._core.create_ufunc` or updating the language standard in `cupy._core.create_ufunc` across the board. There are a number functions in or soon to be in xsf that have not been to CuPy yet and which make use of C++17 features, so this is needed for more than just `invoke_result`.
* `config.h` was previously relying on a private symbol defined by `libcudacxx` to determine if compilation was being done with the CUDA jit `NVRTC`. This symbol was removed in a recent version of `libcudacxx`. Changed to use `__CUDACC_RTC__`.

I've also added a proof of concept Python test file for testing xsf in CuPy. It would be nice configure Pixi to allow for installing the dependencies and running these tests. It would also be nice to set up GPU CI using this test file. Finishing this work would close https://github.com/scipy/xsf/issues/37.

### DIFF
--- a/include/xsf/config.h
+++ b/include/xsf/config.h
@@ -210,8 +210,13 @@ using enable_if = cuda::std::enable_if<Cond, T>;
 template <typename T>
 using decay = cuda::std::decay<T>;
 
-template <typename T>
-using invoke_result = cuda::std::invoke_result<T>;
+template <typename F>
+struct invoke_result {
+    using type = decltype(cuda::std::declval<F>()());
+};
+
+template <typename F>
+using invoke_result_t = typename invoke_result<F>::type;
 
 template <typename T1, typename T2>
 using pair = cuda::std::pair<T1, T2>;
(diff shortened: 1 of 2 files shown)


steppi:
I set up pixi to install dependencies and run the CuPy GPU tests and it works locally. I added a workflow file based on https://github.com/scipy/scipy/blob/main/.github/workflows/gpu-ci.yml, but it seems that the tests aren't getting picked up.

rgommers:
@steppi I fixed an org-level configuration. The repos that the GPU runners are allowed to run on are whitelisted, it's `scipy/scipy` and `scipy/xsf` now. Please try again.

steppi:
Beautiful! CuPy tests running and passing. Thanks @rgommers!

Issue #16, opened 2025-03-12
Errors compiling the library


ZappD0S:
Hi,

I'm trying to compile this library and I'm having a few issues.

- I get an error saying that the function `sf_error()` is not defined. This seems to be called many times throughout the codebase. There is however  another function called `set_error()` that is called in exactly the same way and is defined. I have the impression `set_error()` is the right function that should be used instead. Maybe this was the result of an automated refactoring gone wrong?

- The constant `NPY_LOGE2` referenced in `stats.h` doesn't seem to be defined anywhere.

Disclaimer: I know next to nothing about this library, so maybe the errors I'm getting are not relevant because I'm misusing it. Please be kind!

Thanks a lot.

steppi:
Hi @ZappD0S! Thanks for the issue, you're not doing anything wrong, the feedback is very helpful, and it's exciting your interested in using the library. `xsf` is  a work in progress, and at the moment this code is only tested within SciPy, so there's a good chance that there are little things that won't work outside of SciPy's specific build. For instance `sf_error` was the name of the function before the migration of special function scalar kernel code into `xsf`, and it seems we missed changing to the new name in some cases. This still works in SciPy though, because `sf_error` is still included for scalar kernels that haven't been migrated to `xsf` yet. Similarly, `NPY_LOGE2`, is defined when these headers are included in SciPy, because the NumPy math headers are included in SciPy but not here.

I've been setting up actual CI for `xsf` (https://github.com/scipy/xsf/pull/14) and should have that wrapped up next week. Things like you've noticed should break CI, and I'll try to fix things like that.

Can you let me know which headers you're trying to include in your project? I can priortize getting those fixed for you. Also, since things aren't documented yet, let me know if you have any questions about getting things set up.

ZappD0S:
Hi @steppi, thanks a lot for the detailed reply! I was using only the `stats.h` header.

steppi:
Hi @ZappD0S. I merged https://github.com/scipy/xsf/pull/14 adding CI for `xsf` and fixed the issues you noticed in order to get CI to pass. Hopefully everything should compile for you now.

ZappD0S:
Yes, it works perfectly now. Thanks a lot again!

PR #51, opened 2025-07-23
Fix sf_error typo

SabrinaJewson:
I’m getting the same errors in #16, and it seems to be due to a typo. Just sending a quick fix, but it might point to deeper problems with CI.

### DIFF
--- a/include/xsf/cephes/hyperg.h
+++ b/include/xsf/cephes/hyperg.h
@@ -270,7 +270,7 @@ namespace cephes {
 
             while (t > MACHEP) {
                 if (bn == 0) { /* check bn first since if both   */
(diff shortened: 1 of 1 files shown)


steppi:
Thanks @SabrinaJewson! You're right, the deeper problem is that only code which is actually used in SciPy or CuPy is currently tested. Prior to the separation of `xsf` into a separate library, SciPy began using the Boost implementation of the confluent hypergeometric function `hyp1f1` in place of the Cephes implementation. So the Cephes implementation isn't tested in this repo. `sf_error` is an error handling function internal to SciPy's codebase, but it should have been replaced with `set_error` when this code was moved to `xsf`.

PR #56, opened 2025-08-30
Fix C++ std::complex template compilation error

Emiliano-DM:
Fix C++ std::complex template compilation error in AMOS special functions library. Resolves expected unqualified-id before '_Complex' build failure that prevents SciPy from compiling when
   using C++ std::complex templates.

  Problem

  The C99 complex.h header defines a macro #define complex _Complex which breaks C++ std::complex template usage:

  // C99 complex.h defines:
  #define complex _Complex

  // This breaks C++ code:
  std::complex<double> z;  // Becomes: std::_Complex<double> z; (invalid syntax)

  Build Error:
  error: expected unqualified-id before '_Complex'
  std::complex<double> z;

  Root Cause

  When #include <complex.h> is processed in C++ compilation context, the complex macro interferes with std::complex template instantiation, causing syntax errors in functions that use
  Bessel and Airy function computations.

  Solution

  Add C++ compatibility guard to safely undefine the problematic macro in C++ contexts while preserving C functionality:

  #include <complex.h>
  #ifdef __cplusplus
  #undef complex  // Remove C macro that breaks C++ std::complex
  #endif
  #include <math.h>

  Technical Details

  Standards Compliance:
  - ✅ Maintains C99 complex number functionality
  - ✅ Enables C++ std::complex template usage
  - ✅ Cross-platform compatibility (GCC, Clang, MSVC)

  Industry Pattern:
  - Follows identical approach used by NumPy 2.0
  - Matches Boost.Math library compatibility solutions
  - Standard practice for C/C++ interoperability in scientific libraries

  Files Modified

  - include/xsf/amos/amos.h: Added 3-line C++ compatibility guard

  Testing

  Verified with:
  - GCC 11.2.0 on Linux x86_64
  - C++17 standard compilation
  - Integration with SciPy build system (meson/ninja)

  Compilation Test:
  g++ -I./include -std=c++17 -c test_amos_cpp.cpp  # ✅ Success

  Impact

  Scope: Affects only C++ compilation contexts - zero impact on C code functionality

  Risk Level: Minimal - defensive programming with feature guards


lucascolley:
do you have a build log that fails without this change?

Emiliano-DM:
I have logs with the errors in building yes. The line 104 |     int acai(std::complex<double>, double, int, int, int, std::complex<double> *, double, double, double, double);
[original_state_build_sanitized.log](https://github.com/user-attachments/files/22057914/original_state_build_sanitized.log)

WarrenWeckesser:
I probably won't be able to check this carefully in the next few days, so I'll just add a general warning that undefining macros that come from a standard header in a header that might be included by other code (including downstream users of xsf) is a *very bad practice*.  Don't break public headers!

Having said that, I also realize there have been persistent problems of writing generic code with complex numbers, although the context has usually been Microsoft's inability to actual implement a compliant complex type in their *C* compiler.

WarrenWeckesser:
Can we instead fix `amos.h` to not include the *C* header file `complex.h`? That is, change `#include <complex.h>` to `#include <complex>`, and then fix whatever needs to be fixed in the file. I don't know if there are any obstacles to doing so, but that would seem to be the preferred approach.

PR #58, opened 2025-08-30
Fix C++ std::complex compatibility using standard C++ headers

Emiliano-DM:
## Summary

  Fix C++ template compilation errors when including `amos.h` from C++ code by using standard C++ headers instead of problematic C headers.

  ## Problem

  **Error:** `error: expected unqualified-id before '_Complex'` in C++ files including amos.h

  **Root Cause:**
  - `#include <complex.h>` defines macro `#define complex _Complex`
  - This breaks C++ template syntax: `std::complex<double>` becomes `std::_Complex<double>` (invalid)

  **Example failure:**
  ```cpp
  // amos.h function declaration:
  int acai(std::complex<double>, ...);  // Becomes: int acai(std::_Complex<double>, ...); ❌

  Solution

  Clean C++ standards approach:
  - ✅ Replace #include <complex.h> with #include <complex>
  - ✅ Replace conj() with std::conj() (17 instances)

  Benefits

  - ✅ Eliminates macro pollution - no more #define complex _Complex
  - ✅ Standard C++ practice - uses proper <complex> header
  - ✅ No downstream side effects - safe for code including this header
  - ✅ Functionally identical - std::conj() behaves exactly like conj()
  - ✅ Addresses reviewer concerns - avoids undefining standard macros in public headers

  Changes

  File: include/xsf/amos/amos.h
  - Line 97: #include <complex.h> → #include <complex>
  - 17 locations: conj(...) → std::conj(...)

  Testing

  ✅ Direct C++ compilation test confirms template errors resolved✅ No functional changes - purely implementation improvement✅ Maintains full compatibility with existing function
  signatures

### DIFF
--- a/include/xsf/amos/amos.h
+++ b/include/xsf/amos/amos.h
@@ -94,7 +94,7 @@
 
 #include <stdlib.h>
 
-#include <complex.h>
+#include <complex>
 #include <math.h>
 #include <memory> // unique_ptr
 
@@ -2025,8 +2025,8 @@ namespace amos {
         zn = -z * ci;
         if (yy < 0.0) {
             zn = -zn;
-            csgn = conj(csgn);
-            ci = conj(ci);
+            csgn = std::conj(csgn);
+            ci = std::conj(ci);
         }
         nz = binu(zn, fnu, kode, n, cy, rl, fnul, tol, elim, alim);
         if (nz < 0) {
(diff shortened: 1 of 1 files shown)


WarrenWeckesser (approved):
Looks good to me.

steppi:
Thanks @Emiliano-DM. For context, what happened here is that AMOS was translated from C to C++ very quickly to address the Windows build failures here https://github.com/scipy/scipy/issues/20291 just before the 1.13 release. Basically, we went with the minimal diff that made the current test suite pass, with the idea that we'd come back and clean things up later. Thanks @WarrenWeckesser for reviewing.

PR #62, opened 2025-09-09
BUG: missing complex ``lqmn`` derivatives when ``|z|<1.001``

jorenham:
Closes https://github.com/scipy/scipy/issues/23589

### DIFF
--- a/tests/scipy_special_tests/test_lqmn.cpp
+++ b/tests/scipy_special_tests/test_lqmn.cpp
@@ -6,6 +6,11 @@
 #define MDSPAN_USE_PAREN_OPERATOR 1
 #include <xsf/third_party/kokkos/mdspan.hpp>
 
+// Type aliases for commonly used mdspan types
+using mdspan_1d_double = std::mdspan<double, std::dextents<ptrdiff_t, 1>>;
+using mdspan_2d_double = std::mdspan<double, std::dextents<ptrdiff_t, 2>>;
+using mdspan_2d_cdouble = std::mdspan<std::complex<double>, std::dextents<ptrdiff_t, 2>>;
+
 // From https://github.com/scipy/scipy/blob/bdd3b0e/scipy/special/tests/test_legendre.py#L693-L697
 TEST_CASE("lqmn TestLegendreFunctions.test_lqmn", "[lqmn][lqn][real][smoketest]") {
     constexpr double atol = 1.5e-4;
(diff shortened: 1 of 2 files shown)


jorenham:
My VSCode autoformatter was a bit too enthusiastic, but it should be fixed now

jorenham:
Erm, I can't seem to find the `lqmn` tests; are there any?

jorenham:
> I have no idea what the macos and windows test failures are about though (first time writing non-trivial C++) 😅. Could it be a caching thing, or did I mess up some obvious C pointer thingy or something? Is there a way for me to repro this on my linux (ubuntu) machine? 

I managed to figure it out myself, and it turns out that `lqmn` apparently indexes the first 2x2 indices of the output matrices, even if it's smaller (`m < 1 || n < 1`) than that ([here](https://github.com/jorenham/xsf/blob/4929f4ace18c69d47bd07fd11439d091805ea69f/include/xsf/legendre.h#L861-L869) and [here](https://github.com/jorenham/xsf/blob/4929f4ace18c69d47bd07fd11439d091805ea69f/include/xsf/legendre.h#L983-L994)). On windows this was causing a build failure (stricter compiler or something), and on macos this was causing a runtime crash (without reporting an error code or something). No idea why it was running without problems on linux though.

Either way, I worked around the by increasing the buffer size accordingly if needed, and side-stepped the windows build failure by using 🦆-typed `std::mdspan` classes. 

But as far as I could tell, this `lqmn` indexing issue is also present in the original Fortran code ([here](https://github.com/scipy/scipy/blob/5e4a5e3785f79dd4e8930eed883da89958860db2/scipy/special/specfun/specfun.f#L203-L210) and [here](https://github.com/scipy/scipy/blob/5e4a5e3785f79dd4e8930eed883da89958860db2/scipy/special/specfun/specfun.f#L11574-L11587)). So if this is actually an issue that should be fixed, doing so in a separate PR would have my preference, since it's independent of the one addressed in this PR.

steppi:
Nice work @jorenham! Rather than fixing `lqmn`, what I want to do is push forward with rewriting `lqmn` in the style @izaid used for `assoc_legendre_p` and `assoc_legendre_p_all`. I want to get the proper tests set up first though.

tests/scipy_special_tests/test_lqmn.cpp
```diff
@@ -0,0 +1,152 @@
+#include "../testing_utils.h"
+
+#include <xsf/legendre.h>
+
+// `std::mdspan` (the kokkos one) was causing issues on windows and mac, so we duck-type it instead
```
steppi, line 5: Can you give me info on what was happening here? `std::mdspan` works on all of these platforms when this code is used in SciPy. I think we should take time to figure out why this wasn't working.

jorenham: Ah that comment is outdated, gimme a sec

jorenham: Actually, I'll try it with `mdspan` again, since other things have changed now as well.

jorenham: I'm guessing that it has to do with those template `const` thingies, so that the the compiler detects the OOB indexing. 

So I guess the cleanest approach would be to also use larger `mdspan` extents, matching the buffer.

jorenham:
The windows tests are failing "in the right way" now 😛

steppi (approved):
Thanks for adding the tests @jorenham. I'm a little curious why the existing SciPy tests use an `atol` and not an `rtol` but think it's fine to just copy them over here like this. I'll look into having more systematic tests for legendre functions and other gufuncs.

steppi:
> The windows tests are failing "in the right way" now 😛

Hoping to have a time to investigate fixing the underlying issue soon. It's due to an inconsistency in complex arithmetic on Windows. If I remember right, something to do with them not completely following the spec.

PR #66, opened 2025-10-07
BUG: expi returns NaN upon calling sc.expi(inf)

ChrisAB:
Closes scipy/scipy#23738

### DIFF
--- a/include/xsf/expint.h
+++ b/include/xsf/expint.h
@@ -149,9 +149,9 @@ XSF_HOST_DEVICE inline double expi(double x) {
             }
         }
         ei = ga + std::log(x) + x * ei;
-    } else if (x == HUGE_VAL) {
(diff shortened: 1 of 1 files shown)


include/xsf/expint.h
```diff
@@ -149,6 +149,9 @@ XSF_HOST_DEVICE inline double expi(double x) {
             }
         }
         ei = ga + std::log(x) + x * ei;
+    } else if (x == HUGE_VAL) {
+        // Special use-case needed because exp(inf) / inf is undefined/NaN
+        return HUGE_VAL;
```
steppi, line 154:
```suggestion
    } else if (std::isinf(x)) {
        // Special use-case needed because exp(inf) / inf is undefined/NaN
        return std::numeric_limits<double>::infinity();
```

For consistency with existing code.

Issue #46, opened 2025-07-13
BUG: Incorrect translation of AMOS Fortran code resulting in incorrect/NaN outputs


hchau630:
This following [line](https://github.com/scipy/xsf/blob/2f145ae2ab215dd9702f14921a3450a84e5b98e7/include/xsf/amos/amos.h#L4135C47-L4135C111) in the function `kscl` in `amos.h` is incorrectly translated from the original Fortran code:
```
cs = (exp(std::real(cs)) / tol) * (cos(std::imag(cs)) + sin(std::imag(cs) * std::complex<double>(0, 1)));
```
The original Fortran code (in `zkscl.f`) is:
```
STR = DEXP(CSR)/TOL
CSR = STR*DCOS(CSI)
CSI = STR*DSIN(CSI)
```
so really it should be translated as
```
cs = (exp(std::real(cs))/tol)*std::complex<double>(cos(std::imag(cs)), sin(std::imag(cs)));
```
This resulted in incorrect/NaN outputs for `special.kv` on certain edge case values. For example
```
>>> from scipy import special
>>> special.kv(0, 680-1000j)
np.complex128(nan+infj)
>>> special.kv(0, 680-680j)
np.complex128(-4.5537300329442675e-298-0.02992162857456993j)
```
when really the outputs should be `1.9017e-298+1.7134e-297j` and `-4.5537e-298-1.8787e-297` respectively.

hchau630:
Sorry probably not, since presumably I'll also have to add some new test cases for the fix and I'm not really familiar with the testing pipeline of xsf. Thanks in advance for fixing the issue.

PR #69, opened 2025-10-29
BUG: Fix mistranslated expression in amos.h

WarrenWeckesser:
Closes gh-46.

Thanks to github user hchau630 for identifying the problem and suggesting the fix.


WarrenWeckesser:
I added unit tests.  Since these are new tests that are not related to old scipy tests, I put the test file in a new subdirectory, `xsf_tests` (instead of `scipy_special_tests`).

WarrenWeckesser:
Who else has commit rights to the `xsf` repo?

PR #76, opened 2025-11-18
Bugfix/Workaround sign error Asso. Legendre Function at z=-1

mberz:
This might fix the bug reported in https://github.com/scipy/scipy/issues/23101.

It seems that 
https://github.com/scipy/xsf/blob/15ffa027ca3edc32e8d7a7456ddd360109385acb/include/xsf/legendre.h#L313 should depend on the sign of z.

This is an edge case for abs(z)=1, imag(z)=0 (see https://github.com/scipy/xsf/blob/15ffa027ca3edc32e8d7a7456ddd360109385acb/include/xsf/legendre.h#L402) 

I've implemented a test which catches the original issue in https://github.com/scipy/scipy/pull/24017.

I am not sure if this is a proper solution to the problem or only a workaround. I did not find references in the code of the Legendre functions which were used for the original implementation.

### DIFF
--- a/include/xsf/legendre.h
+++ b/include/xsf/legendre.h
@@ -301,11 +301,11 @@ struct assoc_legendre_p_recurrence_n<T, assoc_legendre_norm_policy> {
 template <typename NormPolicy, typename T>
 void assoc_legendre_p_pm1(NormPolicy norm, int n, int m, T z, int branch_cut, T &res) {
     if (m == 0) {
-	if (real(z) >= 0) {
-	    res = T(1);
-	} else {
-	    res = T(std::pow(-1, n));
-	}
+        if (real(z) >= 0) {
+            res = T(1);
+        } else {
+            res = T(std::pow(-1, n));
+        }
     } else {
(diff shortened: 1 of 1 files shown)


steppi (changes requested):
Nice find @mberz. It looks like it didn't have to do with applying branch cut logic to cases without branch cuts at all, but just a bug in `assoc_legendre_p_pm1`. As it stands, the current patch isn't quite right, but it looks like there's a quick fix fortunately.

steppi:
@mberz, since I haven't left you much time to work on this before SciPy branches for 1.17, I can take it from here.

steppi:
There should be an option to skip the GPU tests for changes that don't impact anything supported on the GPU, or perhaps we could intelligently identify such situations and only run those tests when needed. For now, I just cancelled them manually.

steppi (approved):
OK. I've tweaked the fix and applied it to both overloads where it's needed and have added some relevant tests. All failures are known unrelated failures. I think this is ready now.

PR #79, opened 2025-12-08
BUG(cephes): tukeylambdacdf: fix precision issues

JacobHass8:
This fixes the bug documented in issue [scipy/#21370](https://github.com/scipy/scipy/issues/21370). For $\lambda < 0$ and large $x$, the `stats.tukeylambda.cdf` function would not return 0. By reducing the parameter `tukey_EPS`, this can be achieved.

### DIFF
--- a/include/xsf/cephes/tukey.h
+++ b/include/xsf/cephes/tukey.h
@@ -21,7 +21,7 @@ namespace cephes {
     namespace detail {
 
         constexpr double tukey_SMALLVAL = 1e-4;
-        constexpr double tukey_EPS = 1.0e-14;
+        constexpr double tukey_EPS = 1.0e-18;
         constexpr int tukey_MAXCOUNT = 60;
 
     } // namespace detail
--- a/include/xsf/cephes/tukey.h
+++ b/include/xsf/cephes/tukey.h
@@ -21,7 +21,7 @@ namespace cephes {
     namespace detail {
 
         constexpr double tukey_SMALLVAL = 1e-4;
-        constexpr double tukey_EPS = 1.0e-18;
+        double tukey_EPS = 1.0e-14;
         constexpr int tukey_MAXCOUNT = 60;
 
     } // namespace detail
@@ -42,6 +42,10 @@ namespace cephes {
             if (x >= xeval) {
                 return 1.0;
             }
+        }
+
+        if (x > 1e9){
+            detail::tukey_EPS = 1.0e-18;
         }
 
         if ((-detail::tukey_SMALLVAL < lmbda) && (lmbda < detail::tukey_SMALLVAL)) {


mdhaber:
Thanks. Based on the original report in scipy/scipy#21370, please add a test to this class:
https://github.com/scipy/scipy/blob/b1296b9b4393e251511fe8fdd3e58c22a1124899/scipy/stats/tests/test_distributions.py#L8371
and confirm that it fails before the fix and passes after.

Maybe name it `test_gh21370_cdf_sf_saturation` and as the first line, include a comment summarizing the issue and what the test checks. I don't foresee a need for `parameterize`; you can use an array for `x` (and `lam` if necessary), and tests of the CDF and SF can go on separate lines.

mdhaber:
Oh, I didn't notice this was in xsf and not SciPy. @steppi  can guide you on this one. After it's merged, I think you'll want to open a PR in SciPy that updates the xsf submodule commit and adds the test.

steppi:
I think I need to take a closer look at this. I'm not confident at a glance that changing the value of `tukey_EPS` won't have other consequences. Also, @dschmitz89, is this something that SciPy could use Boost for instead?

WarrenWeckesser:
I worked on `tukeylambda` last year, but it turned into a bit of a can of worms.  I'll get back to it, and continue the discussion over in https://github.com/scipy/scipy/issues/21370.

dschmitz89:
Could you add tests here @JacobHass8 ?

JacobHass8:
> Could you add tests here @JacobHass8 ?

I added tests to the scipy repository here https://github.com/scipy/scipy/pull/24167. They are currently marked as failing, but  pass if this change is made. 

It might be nice to add this distribution to boost/math in the long term.

dschmitz89:
@JacobHass8 Since the goal is to use xsf kernels also in Cupy we prefer to do the tests in this repo directly.

lucascolley:
gentle bump @JacobHass8

JacobHass8:
> gentle bump @JacobHass8

Sorry for the delay! I converted the tests into a pretty simple cpp file using the existing xsf testing framework. However, I'm not sure what to do so the file is run using `pixi run tests`. Do you have any guidance?

dschmitz89:
> > gentle bump @JacobHass8
> 
> Sorry for the delay! I converted the tests into a pretty simple cpp file using the existing xsf testing framework. However, I'm not sure what to do so the file is run using `pixi run tests`. Do you have any guidance?

That sounds right. You can also pass arguments to ctest from `pixi run tests`.

lucascolley:
> I'm not sure what to do so the file is run using pixi run tests. Do you have any guidance?

just put it in a file in the same directory as the other test files, it will be discovered automatically

JacobHass8:
I just needed to clear my cache and it discovered the test file. Thanks for the help! I only tested that the CDF was monotonic but nothing with the SF since the SF isn't implemented in xsf. I assume scipy just returns 1-CDF. 

Edit: Looks like there are formatting errors. I'll fix that tonight.

JacobHass8:
> > I'm not sure what to do so the file is run using pixi run tests. Do you have any guidance?
> 
> just put it in a file in the same directory as the other test files, it will be discovered automatically

Should be good to go now!

dschmitz89:
@JacobHass8 Could you produce a plot of the tukeylambda CDF with logarithmized x scale? I want to know if it is monotonic. The fix here still feels like a "hack" but might be good enough.

JacobHass8:
> @JacobHass8 Could you produce a plot of the tukeylambda CDF with logarithmized x scale? I want to know if it is monotonic. The fix here still feels like a "hack" but might be good enough.

Yes, it's still not monotonic but goes to 0 as x-> inf as it should. I honestly think the solution is just to use a precision of 1e-18 everywhere rather than implementing a cutoff at 1e7 (as in the graph below). The same issue occurs for a cutoff at 1e9 as in the PR. 

image

JacobHass8:
This is for a precision of 1e-18 everywhere. Here, 1-CDF is 0 for all x > ~1e10. 

image

dschmitz89:
@JacobHass8 The second plot looks convincing, we just need to make sure that it does not lead to different issues for small `x` for performance degredations.

PR #82, opened 2025-12-10
Fix bug for special.lpmv(0,v,-1)

fbourgey:
Fixes bug in https://github.com/scipy/scipy/issues/20902


fbourgey:
do we want a test as well?

steppi:
Usually I cut an `xsf` release and then create a PR to update `xsf`.

PR #86, opened 2025-12-20
BUG: amos/amos.h: fix pointer dereference in uni2

stratakis:
The code was performing pointer arithmetic instead of incrementing the value, unlike the identical pattern in uni1.

Found by coverity static analysis.


WarrenWeckesser (approved):
Thanks, that is clearly a bug, and the fix looks good.

steppi (approved):
Thanks @stratakis and @WarrenWeckesser for the review.

PR #88, opened 2025-12-23
ENH: Add negative m for Jacobi elliptic functions

fbourgey:
Allow for negative $m$ in special.ellipj following https://github.com/scipy/scipy/issues/21415

Following https://dlmf.nist.gov/22.17, write $\mathrm{sn}(u|m):=\mathrm{sn}(u,k)$ with $m=k^2$.

If $m<0$, write $m = -\kappa^2$ with $k= \mathrm{i} \kappa$.

$k_1 = \frac{\kappa}{\sqrt{1+\kappa^2}} = \sqrt{m_1}$ where $m_1 := \frac{-m}{1-m} \in (0,1)$ 

and 

$k_1^\prime = \frac{1}{\sqrt{1-m}}$. 

We can then use (22.17.6-7-8) and get

$$
\mathrm{sn}(u|m) = \mathrm{sn}(u,\mathrm{i} \kappa) = \frac{1}{\sqrt{1-m}} \mathrm{sd}(\sqrt{1-m}u| m_1)
$$

$$
\mathrm{cn}(u|m) = \mathrm{cd}(\sqrt{1-m}u| m_1)
$$

$$
\mathrm{dn}(u|m) = \mathrm{nd}(\sqrt{1-m}u| m_1)
$$


dschmitz89 (approved):
LGTM. Is there anything we should have a close look before merging @fbourgey ?

fbourgey:
I think I am getting wrong values for very large negative $m$ (eg, m=-1e10, -1e12). We'll need to update the description (line 23) in `xsf/cephes/ellpj.h` too.

dschmitz89:
> I think I am getting wrong values for very large negative m (eg, m=-1e10, -1e12). We'll need to update the description (line 23) in `xsf/cephes/ellpj.h` too.

I am all for maximum correctness but it this PR not a big improvement over the status quo? Many of our special functions are not accurate for extreme inputs, so I do not think we have to block here. Or are those inputs typical for elliptic functions?

fbourgey:
> > I think I am getting wrong values for very large negative m (eg, m=-1e10, -1e12). We'll need to update the description (line 23) in `xsf/cephes/ellpj.h` too.
> 
> I am all for maximum correctness but it this PR not a big improvement over the status quo? Many of our special functions are not accurate for extreme inputs, so I do not think we have to block here. Or are those inputs typical for elliptic functions?

Yes, this is already a good update and those are very extreme values. I updated the comment.

PR #91, opened 2026-01-28
BUG: specfun/specfun.h: initialize kd in mtu12 to fix potential uninitialized use

stratakis:
The kd variable was not initialized before the kf==1/kf==2 conditional blocks. If kf had an unexpected value, kd would be used uninitialized when passed to cva2(). Initialize to 0 for consistency with mtu0.

Uncovered by Coverity static analysis.


dschmitz89:
Sry that it took so long to take a look @stratakis . Could you rebase on the main branch? If CI comes back green, I will merge here. Coverity has proven itself very useful.

stratakis:
Rebased.

Issue #50, opened 2025-07-22
ENH(bessel): add vectorization of orders


JaRoSchm:
Hi everyone,

In https://github.com/scipy/scipy/issues/19079 I suggested to surface the vectorization capabilities of amos to scipy. In this context vectorization means that amos can return the value of the Bessel functions for several consecutive orders at once as they are computed recursively and the intermediate results are stored. At the moment xsf and scipy don’t allow users to access this functionality. Now I started to work on this functionality and would like to get some early feedback, especially as the changes in xsf and scipy have to be coordinated. The motivation for our research group and me is the computation of scattering resonances, where many consecutive orders of Bessel functions with the same argument are needed. This is also why I chose `hankel1` to start with and which I will use below for some early performance assessments.

In the coming week and months the time I can put into this shouldn’t be the limiting quantity. I have some years of experience of (scientific) python, but I only used C on a basic level. So I am able to produce running C++ code, but may need help for design choices to produce idiomatic code.

# Motivation

Using my fork which I will link below, I was able to do a quick benchmark. The parameters I have chosen are representative for our use case.

```python
import numpy as np
from scipy.special import hankel1_all, hankel1
from time import perf_counter

z = 20_000.0 - 0.5j
v_max = 20_300

start = perf_counter()
results_all = hankel1_all(0, z, v_max + 1)
end = perf_counter()
time_all = end - start
print(f"Time for hankel1_all: {time_all * 1000:.2f} ms")

results = np.empty_like(results_all)
start = perf_counter()
for v in range(v_max + 1):
    results[0, v] = hankel1(v, z)
end = perf_counter()
time = end - start
print(f"Time for hankel1: {time * 1000:.2f} ms")
print(f"Speedup: {time / time_all:.1f}x")

# Output:
# Time for hankel1_all: 0.37 ms
# Time for hankel1: 66.19 ms
# Speedup: 178.1x
```

This shows that the vectorization can lead to a significant speedup.

# Implementation in xsf

You can find my current implementation here: https://github.com/JaRoSchm/xsf/tree/bessel_all

[...]

PR #92, opened 2026-02-03
BUG: amos/amos.h: fix scope of `l` in `seri`

JaRoSchm:
Hi, while in the long term I am still interested of getting https://github.com/scipy/xsf/issues/50 into xsf and scipy, I am using my own Python wrapper around your C translation of AMOS in the meantime. I noticed that for some parameters the implementation of `besh` tries to write the result into `cy[-1]`. To find the origin of this I compared the xsf translation to the automatic AMOS translation from https://github.com/jpcima/zbessel. The wrapper I used for this can be found at https://github.com/JaRoSchm/pyamos. There I added a check which allows me to detect if AMOS tries to write out of the range of the result array, see https://github.com/JaRoSchm/pyamos/blob/1cd435a33db41d3bad2448991ca0e5027a87057e/pyamos.cpp#L44C1-L50C10. To see the problem, you can simply run `pixi run python compare_xsf_zbessel.py` there.

The bug I found using this is the scope of the variable `l` in the `seri` function (see the diff of this PR). After the `goto L100` statement (line 4494), the value of `l` from the loop is needed and not the value `l = 3` assigned beforehand. Without the change from this PR, we had `l = 3` in line 4499, which is not the wanted value. The original Fortran code can be found at https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zseri.f#L147.

I thought about adding a test for this, but I am not sure, how to do this. In https://github.com/JaRoSchm/pyamos/blob/main/compare_xsf_zbessel.py I use `n = 260` resulting in a array of this length, which would have to be added as a reference value. Additionally, I actually never used C++ (and its test libraries), so all of this is an extrapolation of my Python and C knowledge.


JaRoSchm:
I added a test to this PR which fails on main and passes here. In this case, I chose not to include 260 reference values but simply compare the "vectorized order" version to the "scalar" one. The "scalar" version is tested separately and should be fine as a reference. Additionally, I now use "test cases", which will allow me to simply add another test for #93, when this is merged.

JaRoSchm:
@fbourgey Could you please have a look at this PR? I think this is similarly straightforward as my previous PRs to fix translation bugs in amos. After this is merged I can simply add a test case to #93 which otherwise is also ready for review.

tests/xsf_tests/test_amos.cpp
```diff
@@ -62,6 +62,36 @@ TEST_CASE("amos besi vectorized", "[amos][xsf_tests]") {
     }
 }
 
+TEST_CASE("amos besh vectorized", "[amos][xsf_tests]") {
```
fbourgey: This test seems useful but should we not add a sentinel regression test as you did https://github.com/scipy/xsf/pull/158? It passes on main.

include/xsf/amos/amos.h
```diff
@@ -4476,8 +4476,8 @@ namespace amos {
             //
             s1 = w[0];
             s2 = w[1];
-            l = 3;
-            for (int l = 3; l < (nn + 1); l++) {
+            int l;
+            for (l = 3; l < (nn + 1); l++) {
```
fbourgey, line 4480: Indeed, this seems to fix the translation error.

tests/xsf_tests/test_amos.cpp
JaRoSchm: Thanks for checking! Seems like I have mixed up what this PR actually fixes with some other PR. I have now added a test similar to #153 here.

tests/xsf_tests/test_amos.cpp
```diff
@@ -62,6 +62,28 @@ TEST_CASE("amos besi vectorized", "[amos][xsf_tests]") {
     }
 }
 
+TEST_CASE("amos seri buffer overflow gh-92", "[amos][xsf_tests]") {
```
fbourgey, line 65: I tried to run this test on `upstream/main` and it still passes.

fbourgey (changes requested):

tests/xsf_tests/test_amos.cpp
JaRoSchm: Hm, I tried again and for me the test fails if I revert the fix. I reverted the fix and pushed to see what CI says.

JaRoSchm: Okay, looks like you are using macOS? Then CI agrees with you. On Linux CI reproduces the failure I observe. So the effect of the bug might be ill-defined in C++ and gcc/clang give different results. Could this be the origin of the differences?

Would you be willing to merge if I revert the last commit? Looks like we agree that the fix is correct and we have a regression test at least for some of the CI jobs.

fbourgey: I think those are memory bugs. I would expect CI to fail for all three Operating Systems.

tests/xsf_tests/test_amos.cpp
```diff
@@ -62,6 +62,28 @@ TEST_CASE("amos besi vectorized", "[amos][xsf_tests]") {
...
+    // allocate n+1 elements, initialize extra to sentinel to detect overflow
+    const complex<double> sentinel{12345.67, 98765.43};
+    std::vector<complex<double>> cy(n + 1, sentinel);
+
+    int ierr = 0;
+    int nz = xsf::amos::besh(z, fnu, kode, m, n, cy.data(), &ierr);
+
+    // check if the extra element (index n) was touched
+    CAPTURE(cy[n]);
+    CHECK(cy[n] == sentinel);
```
fbourgey: @JaRoSchm WDTY? This seems to fail on MacOS.

```suggestion
    const complex<double> sentinel{12345.67, 98765.43};
    std::vector<complex<double>> cy(n + 2, sentinel);

    int ierr = 0;
    int nz = xsf::amos::besh(z, fnu, kode, m, n, cy.data() + 1, &ierr);

    // check if guards elements were touched
    CAPTURE(cy[0], cy[n + 1]);
    CHECK(cy[0] == sentinel);
    CHECK(cy[n + 1] == sentinel);
```

fbourgey (changes requested):

tests/xsf_tests/test_amos.cpp
JaRoSchm: Ah, makes sense. Especially, since I already wrote in the PR description that `cy[-1]` is touched. Thanks!

JaRoSchm:
CI failed successfully, see https://github.com/scipy/xsf/pull/92/checks?sha=8a3cb53b1425a2de1d73dd67fe959bc6513ec017. I put the fix back in. I will update the test for #93 accordingly.

fbourgey:
Thanks!

JaRoSchm:
Thank you for the review and for improving the test!

PR #93, opened 2026-02-03
BUG: amos/amos.h: fix loop upper bound in `buni`

JaRoSchm:
Hi, I detected this error in the same way as in #92. This time I used `z = 90.0 - 3.0j` and `n = 30` in https://github.com/JaRoSchm/pyamos/blob/main/compare_xsf_zbessel.py for finding the bug. This was simply incorrectly translated from FORTRAN, see https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zbuni.f#L114, where only the lower bound has been shifted by one but not the upper bound. As `k` is reduced by one for each loop, it will become -1 during the last iteration, such that `y` as accessed at -1. Regarding testing the same question/comment from #92 applies.

### DIFF
--- a/include/xsf/amos/amos.h
+++ b/include/xsf/amos/amos.h
@@ -3752,7 +3752,7 @@ namespace amos {
             nl = n - 1;
             fnui = nl;
             k = nl;
-            for (i = 0; i < (nl + 1); i++) {
+            for (i = 0; i < nl; i++) {
                 st = s2;
                 s2 = (fnu + fnui) * rz * s2 + s1;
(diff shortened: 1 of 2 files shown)


JaRoSchm:
I checked that CI fails if the fix is reverted, see https://github.com/scipy/xsf/pull/93/commits/ca1dbe87a84691a3bb4353ae708a698699ba1972.

fbourgey:
Thanks @JaRoSchm!

PR #96, opened 2026-02-11
fix complex `sph_bessel_j` kernel `sqrt` branch

lucascolley:
https://github.com/scipy/scipy/pull/23700#issuecomment-3886895326 @steppi 

AI tools used: ChatGPT

The SciPy tests seem to pass on osx-arm64 with this change.

### DIFF
--- a/include/xsf/sph_bessel.h
+++ b/include/xsf/sph_bessel.h
@@ -112,7 +112,7 @@ std::complex<T> sph_bessel_j(long n, std::complex<T> z) {
         return 0;
     }
 
-    std::complex<T> out = std::sqrt(static_cast<T>(M_PI_2)) * cyl_bessel_j(n + static_cast<T>(0.5), z) / std::sqrt(z);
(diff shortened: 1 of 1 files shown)


steppi:
Thanks @lucascolley. I think this will work. See the comment about defining a constant to use here. Also, the fix is needed in the similar line in each of the spherical bessel functions defined in this header.

lucascolley:
Do we have the relevant test coverage in this repo?

lucascolley:
> I found that `xsref` has no tables for the spherical bessel functions. I think this is because the public SciPy functions are regular functions that wrap ufuncs instead of ufuncs themselves. I used a script to automatically generate all of the tables based on what functions get called on what arguments in scipy special's test suite, and private ufuncs with public wrappers only got included if I manually added them.
> 
> 
> 
> It shouldn't be too bad to add the tables. I might have time over the weekend.

gentle bump @steppi

lucascolley:
I believe this has been superseded by gh-151, will see how the linux aarch64 Pixi PR is doing now soon

PR #107, opened 2026-03-13
BUG: lambertw: fix special case `z = -1/e`

ilanschnell:
This PR fixes: https://github.com/scipy/scipy/issues/24770

After inspecting the output of scipy's lambertw function near -1/e+0j, I found that only the branch point -1/e itself is the problem, as it will cause division by zero in Halley's method.  When changing even the least significant (of the real or imaginary part of z), the method converges to the correct result.

### DIFF
--- a/include/xsf/lambertw.h
+++ b/include/xsf/lambertw.h
@@ -94,7 +94,7 @@ XSF_HOST_DEVICE inline std::complex<double> lambertw(std::complex<double> z, lon
         // The imaginary return value is calculated from mpmath using the
         // double precision floating point value -0.36787944117144233,
         // which is slightly less than the mathematical value of -1/e.
-        return std::complex<double>(-1.0, (k ? -1:+1) * 8.220079714836618e-9);
+        return std::complex<double>(-1.0, (k ? -1 : +1) * 8.220079714836618e-9);
     }
 
     absz = std::abs(z);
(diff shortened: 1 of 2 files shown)


WarrenWeckesser:
> `EXPN1` is defined as 0.36787944117144232159553, which by far exceeds the double floating point precision.

It is a double precision floating point constant.  The compiler won't store more than what fits in a double precision float, so writing it that way is no more precise than `-0.36787944117144233`.  You can see the same thing in Python:

```
In [327]: 0.36787944117144232159553
Out[327]: 0.36787944117144233
```
So there shouldn't be any need to change the literal constant or the expression that checks `z == -EXPN1`.  I guess the definition of `EXPN1` could be changed to `0.36787944117144233`, so future readers of the code see a value that is representative of the *actual* value that the compiler will use, but that is not necessary in this PR.

If the literal value ended with the `l` suffix, the compiler would parse it as a `long double`, but there is no point in doing that, since the precision of `long double` is platform-dependent (and is the same as `double` on some platforms), and rest of the variables are just double precision.

ilanschnell:
While working on the new tests, I realized that when lowering the tolerance `lambertw` still gives `nan` in some cases when z is close to -1/e:
```
In [27]: from scipy.special import lambertw

In [28]: lambertw(-0.36787944, -1)
Out[28]: np.complex128(-1.0000000095529311+0j)

In [29]: lambertw(-0.36787944, -1, 1e-12)
Out[29]: np.complex128(nan+nanj)
```
Therefore the comment
```
 * TODO: use a series expansion when extremely close to the branch point
 * at `-1/e` and make sure that the proper branch is chosen there.
```
remains valid,  However, as no `nan`-values are returned when using scipy's default tolerance (1e-8), I consider this less critical now.  I'm happy to work on the series expansion near -1/e in a separate PR once this PR gets merged.

Issue #98, opened 2026-02-14
ENH: Port null-distribution special functions from scipy.stats to xsf


fbourgey:
Following @mdhaber's [comment](https://github.com/scipy/scipy/pull/24020#issuecomment-3887411438) and @steppi's suggestion.

### Motivation
Several null distributions of test statistics are currently implemented in `scipy.stats`, even though they are essentially *special functions*. The issue tracks and coordinates porting these functions to `xsf`.

Moving these functions to `xsf` would:
- make them more reusable by other libraries
- leverage Special’s infrastructure (delegation, `lazy_apply`, etc.)
- enable efficient compiled (C/C++) implementations
- improve performance for repeated CDF/SF evaluations at many statistic values for the same sample size, where Python-level caching is often insufficient

### Proposed starting point
[`_cdf_cvm_inf`](https://github.com/scipy/scipy/blob/6d563be5f49b5d847d20cae1be74c84f7516d1ba/scipy/stats/_hypotests.py#L457): asymptotic (infinite-sample-size) null distribution CDF of the one-sample Cramér–von Mises statistic 
- https://github.com/scipy/xsf/pull/97

This would serve as a pilot for additional ports.

### Other asymptotic or exact null distributions that could follow:
- Cramér–von Mises: [`_cdf_cvm`](https://github.com/scipy/scipy/blob/6d563be5f49b5d847d20cae1be74c84f7516d1ba/scipy/stats/_hypotests.py#L496)
- Ansari
- Wilcoxon
- Mann–Whitney U
- Kendall’s tau
- Kolmogorov–Smirnov

mdhaber:
1. Whatever the implementer and @steppi agree on for starters is fine with me. OK with me to start without a cache and add later, if that's easier.
2. I remember from the initial implementation of MVN/MVT, it would be possible to pass in uniform random data. There is a rejection component, but I think we had a plan for dealing with that. It's OK to leave those at the bottom of the list.
3. Great. I'll add a list.

fbourgey:
For [KS](https://github.com/scipy/scipy/blob/a4463edad21e858fb7556b4fd121463da8da2e33/scipy/stats/_continuous_distns.py#L197), do we port all methods (pdf/cdf/sf/ppf/isf)?

mdhaber:
I guess I only need the ones that are used in the test. Up to Albert whether he wants the rest, but maybe we at least start with the `cdf`s/`sf`s for all the distributions, and come back to other methods later if needed.

[...]

PR #108, opened 2026-03-13
ENH: Implement the exact p-value of the Cramér-von Mises two-sample test

fbourgey:
### Reference issue

Implement the exact p-value of the Cramer-von Mises two-sample test for a given value of the test statistic from [`_pval_cvm_2samp_exact`](https://github.com/scipy/scipy/blob/6d563be5f49b5d847d20cae1be74c84f7516d1ba/scipy/stats/_hypotests.py#L1568)

https://en.wikipedia.org/wiki/Cram%C3%A9r%E2%80%93von_Mises_criterion

### DIFF
--- a/include/xsf/stats.h
+++ b/include/xsf/stats.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "xsf/bessel.h"
+#include "xsf/binom.h"
 #include "xsf/cephes/bdtr.h"
 #include "xsf/cephes/chdtr.h"
 #include "xsf/cephes/fdtr.h"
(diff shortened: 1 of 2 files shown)


include/xsf/stats.h
```diff
@@ -284,6 +285,87 @@ inline double pdtrc(double k, double m) { return cephes::pdtrc(k, m); }
 
 inline double pdtri(int k, double y) { return cephes::pdtri(k, y); }
 
+inline int64_t comb(int64_t n, int64_t k) {
```
mdhaber: Is `binom` really not implemented?

include/xsf/stats.h
```diff
@@ -284,6 +285,87 @@ inline double pdtrc(double k, double m) { return cephes::pdtrc(k, m); }
...
+            int64_t res = diff * diff;
+            // tmp[0] += res  (shift all keys by res)
+            tmp.clear();
+            for (const auto &[key, freq] : merged) {
+                tmp[key + res] += freq;
+            }
+            next_gs.push_back(tmp);
+        }
+        gs = std::move(next_gs);
+    }
```
mdhaber: Can you provide additional information that helps the reader compare this to either the original code, which is pretty hard to follow, or the algorithm, which looks simple but has terms like $I[a^2u(u+1)(2u + 1)/6]$ and $I[b^2u(u+1)(2u + 1)/6]$ that don't obviously appear in either version? If it would be cleaner, feel free to express this in terms of Algorithm 1 from the paper rather than the Python code. There are a lot of aspects of the Python that don't look like a very efficient expression of the algorithm (lots of placeholding size-zero array, use of arrays instead of `dict`s, use of `delete`, etc.), so if it can't be translated directly, it is probably not worth trying to maintain a resemblance.

fbourgey: I agree; I've found the Python code quite difficult to follow. I’ll provide more detail and include screenshots of the relevant sections of the paper in the coming days.

fbourgey:
I am closing this PR as https://github.com/scipy/xsf/pull/118 supersedes it.

PR #110, opened 2026-03-19
ENH(stats): implement Kendall discordance

fbourgey:
### Reference issue

Implement Kendall discordance from [`_kendall_dis`](https://github.com/scipy/scipy/blob/a4463edad21e858fb7556b4fd121463da8da2e33/scipy/stats/_stats.pyx#L84)

### DIFF
--- a/include/xsf/stats.h
+++ b/include/xsf/stats.h
@@ -158,6 +158,34 @@ inline double gdtr(double a, double b, double x) { return cephes::gdtr(a, b, x);
 
 inline double gdtrc(double a, double b, double x) { return cephes::gdtrc(a, b, x); }
 
+inline int64_t kendall_dis(const std::vector<intptr_t> &x, const std::vector<intptr_t> &y) {
+    intptr_t sup = 1 + *std::max_element(y.begin(), y.end());
+    // Use of `>> 14` improves cache performance of the Fenwick tree (see gh-10108)
+    std::vector<intptr_t> arr(sup + ((sup - 1) >> 14), 0);
+    intptr_t i = 0, k = 0, size = static_cast<intptr_t>(x.size()), idx;
+    int64_t dis = 0;
+    while (i < size) {
+        while (k < size and x[i] == x[k]) {
+            dis += i;
+            idx = y[k];
+            while (idx != 0) {
+                dis -= arr[idx + (idx >> 14)];
+                idx = idx & (idx - 1);
+            }
+            ++k;
+        }
+        while (i < k) {
+            idx = y[i];
+            while (idx < sup) {
+                arr[idx + (idx >> 14)] += 1;
+                idx += idx & -idx;
+            }
+            ++i;
+        }
+    }
+    return dis;
+}
+
 inline double kolmogorov(double x) { return cephes::kolmogorov(x); }
 
 inline double kolmogc(double x) { return cephes::kolmogc(x); }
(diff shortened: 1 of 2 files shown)


mdhaber (approved):
This looks like a faithful translation, and the tests are OK with my if they're ok with @steppi. 

The tests are a little light, so I would suggest a few larger, non-trivial examples (probably randomly generated), and also checking behavior when there are tied elements. Maybe run `kendalltau` with randomly generated `x` and `y` (with and without the possibilities of ties in `x`, ties in `y`, and both) up until you get to `dis = _kendall_dis(x, y)  # discordant pairs` to see what sort of input `_kendall_dis` can receive.

Thank you!

@steppi if we use this in SciPy, can it automatically be vectorized to work along the last axis?

fbourgey:
> This looks like a faithful translation, and the tests are OK with my if they're ok with @steppi.
> 
> The tests are a little light, so I would suggest a few larger, non-trivial examples (probably randomly generated), and also checking behavior when there are tied elements. Maybe run `kendalltau` with randomly generated `x` and `y` (with and without the possibilities of ties in `x`, ties in `y`, and both) up until you get to `dis = _kendall_dis(x, y) # discordant pairs` to see what sort of input `_kendall_dis` can receive.
> 
> Thank you!
> 
> @steppi if we use this in SciPy, can it automatically be vectorized to work along the last axis?

Thanks @mdhaber, I'll improve the tests.

fbourgey:
@dschmitz89 when you get the chance, can you please review that one?

include/xsf/stats.h
```diff
@@ -158,6 +158,34 @@ inline double gdtr(double a, double b, double x) { return cephes::gdtr(a, b, x);
 
 inline double gdtrc(double a, double b, double x) { return cephes::gdtrc(a, b, x); }
 
+inline int64_t kendall_dis(const std::vector<intptr_t> &x, const std::vector<intptr_t> &y) {
```
dschmitz89, line 161: Does this need to be CuPy compatible in future? then `std::vector` is a no go from what I understood.

dschmitz89:
This LGTM, only question left is the CuPy compatibility.

include/xsf/stats.h
steppi: Right, and I have https://github.com/cupy/cupy/pull/9959 working now, so it's something we can actually try out now off of that branch.

steppi: This should take three mdspan views, for `x` and `y`, but also for the scratch buffer space `arr`. No allocations inside of kernels. The SciPy gufunc should use the stateful functor pattern from https://github.com/steppi/scipy/blob/759ab1c458d56404dc8938a6074ebec8d290b449/scipy/special/mathieu.h to reuse the scratch buffer over the course of a single gufunc call. The CuPy gufunc will pre-allocate the scratch buffer space in Python wrapper.

include/xsf/stats.h
```diff
@@ -158,6 +158,34 @@ inline double gdtr(double a, double b, double x) { return cephes::gdtr(a, b, x);
...
+    intptr_t sup = 1 + *std::max_element(y.begin(), y.end());
```
steppi, line 162: Wait,  I missed that the size of the scratch buffer depends on the values of `y`. I think the way this is done needs to be rethought. I'll look into it.

steppi: It seems possible to use a different algorithm (Knight's merge-sort based algorithm) that won't require a buffer with size depending on the values of `y`. I'll take this one from here.

PR #111, opened 2026-03-19
ENH: Implement CDF of the von Mises distribution

fbourgey:
### Reference issue

Implement the CDF of the von Mises distribution and related helper functions from [`von_mises_cdf`](https://github.com/scipy/scipy/blob/a4463edad21e858fb7556b4fd121463da8da2e33/scipy/stats/_stats.pyx#L19-L80)


dschmitz89:
The math looks sound here. Input and output should be a simple scalar here so we can turn this into a regular ufunc. I do not understand why this was taking arrays as input before. Could you chime in @WarrenWeckesser ?

WarrenWeckesser:
> Input and output should be a simple scalar here so we can turn this into a regular ufunc. I do not understand why this was taking arrays as input before.

I agree; I would expect the C++ implementation of `von_mises_cdf(k, x)` in `xsf` to be a scalar kernel.  In SciPy, that kernel would be wrapped using the standard ufunc machinery so that the Python function implements broadcasting and the rest of the ufunc API (including, for example, the `out` parameter).

The existing Cython code accepts input arrays with arbitrary dimensions, but it uses NumPy functions and loops to create an *ad hoc* implementation of broadcasting of `k` and `x`.

steppi:
@fbourgey, to help you keep on track, I recommend pairing each xsf PR adding a new function to an associated SciPy PR adding a new ufunc/gufunc (temporarily checking out and pinning the xsf subproject to your xsf PR branch). This is the workflow I use.

fbourgey:
Thanks all. I will make this a scalar kernel ASAP.

@steppi regarding your [comment](https://github.com/scipy/xsf/pull/111#issuecomment-4407129456), should I go ahead and submit a PR on SciPy for https://github.com/scipy/xsf/pull/130?

steppi:
> Thanks all. I will make this a scalar kernel ASAP.
> 
> @steppi regarding your [comment](https://github.com/scipy/xsf/pull/111#issuecomment-4407129456), should I go ahead and submit a PR on SciPy for #130?

Yeah, when it's ready. Usually what I try to do is work on SciPy and xsf feature branches in parallel and PR them at the same time when they are ready for review. Then we can merge the xsf PR first, make an xsf release, and then change the xsf submodule in the SciPy from your associated xsf feature branch to the new release.

fbourgey:
Sounds good. https://github.com/scipy/xsf/pull/130 has been merged already but maybe we can merge https://github.com/scipy/xsf/pull/112 and https://github.com/scipy/xsf/pull/111 before making a new xsf release?

fbourgey:
@dschmitz89 I've tried to make this a scalar kernel and added a few tests with some SciPy reference values.

include/xsf/stats.h
```diff
@@ -300,4 +301,55 @@ inline float tukeylambdacdf(float x, double lmbda) {
...
 
+XSF_HOST_DEVICE inline double von_mises_cdf_series(double k, double x, unsigned int p) {
+    double s, c, sn, cn, R, V;
+    unsigned int n;
+    s = std::sin(x);
+    c = std::cos(x);
+    sn = std::sin(p * x);
+    cn = std::cos(p * x);
+    R = 0;
+    V = 0;
```
dschmitz89: I guess you just followed the code in SciPy but let's use typical C++ style and use lower letters for variables.

dschmitz89 (changes requested):
I think we can get this into the next xsf release once the merge conflicts are resolved.

My main blocker is adding references: does the current SciPy code or the PR that added this include some useful ones?

fbourgey:
> I think we can get this into the next xsf release once the merge conflicts are resolved.
> 
> My main blocker is adding references: does the current SciPy code or the PR that added this include some useful ones?

@dschmitz89 I've used lower case letters, strengthen test coverage. I've added the same tests as those in [scipy/stats/tests/test_distributions.py::TestVonMises](https://github.com/scipy/scipy/blob/main/scipy/stats/tests/test_distributions.py#L79). Let me know what you think.

dschmitz89:
Thanks @fbourgey , tests and implementation look good! I would still like to add a reference to the code comments.

fbourgey:
> Thanks @fbourgey , tests and implementation look good! I would still like to add a reference to the code comments.

Sorry, I missed that. I've just added a ref.

include/xsf/stats.h
```diff
@@ -568,4 +569,69 @@ XSF_HOST_DEVICE inline typename InputMat::value_type take_from_discrete_cdf(Inpu
...
+            sn = sn_new;
+            cn = cn_new;
+            r = k / (2 * n + k * r);
+            v = r * (sn / n + v);
+        }
+        return 0.5 + x / (2.0 * M_PI) + v / M_PI;
+    }
+
+    XSF_HOST_DEVICE inline double von_mises_cdf_normalapprox(double k, double x) {
+        double b = xsf::cephes::detail::SQRT2OPI / cephes::i0e(k); // Check for negative k
```
j-bowhay: Its not entirely clear what the "check for negative k" comment means?

tests/xsf_tests/test_von_mises_cdf.cpp
```diff
@@ -0,0 +1,150 @@
+#include "../testing_utils.h"
+#include <cmath>
+#include <xsf/stats.h>
+
+TEST_CASE("von Mises CDF test", "[von_mises_cdf][xsf_tests]") {
+    // Reference values from scipy.stats._stats
```
j-bowhay: It might be good to include the permalink since presumably `von_mises_cdf` will be removed at some point if it is replaced by this

dschmitz89 (approved):

fbourgey:
@dschmitz89 @j-bowhay do you think this can be merged now?

PR #112, opened 2026-03-20
ENH: Implement CDF of the bivariate normal distribution

fbourgey:
### Reference issue

Implement the CDF of the bivariate normal distribution from [`_bvnu`](https://github.com/scipy/scipy/blob/a4463edad21e858fb7556b4fd121463da8da2e33/scipy/stats/_stats_pythran.py#L226)

### Additional information

I have kept the original comments (MATLAB code) from [`_bvnu`](https://github.com/scipy/scipy/blob/a4463edad21e858fb7556b4fd121463da8da2e33/scipy/stats/_stats_pythran.py#L226)


include/xsf/stats.h
```diff
@@ -268,6 +268,187 @@ inline std::complex<float> log_ndtr(std::complex<float> z) {
...
+        } else {
+            p = ndtr(-dk);
+        }
+    } else if (dk == -math_inf) {
+        // elseif dk == -inf, p = phid(-dh);
+        p = ndtr(-dh);
+    } else if (r == 0.0) {
+        // elseif r == 0, p = phid(-dh)*phid(-dk);
+        p = ndtr(-dh) * ndtr(-dk);
+    } else {
```
dschmitz89: If this part is factored out into its own function we can reduce a full level of indentation for the rest of the function. WDYT?

include/xsf/stats.h
```diff
@@ -268,6 +268,187 @@ inline std::complex<float> log_ndtr(std::complex<float> z) {
...
+        // else, tp = 2*pi; h = dh; k = dk; hk = h*k; bvn = 0;
+        double tp = 2 * M_PI;
+        double h = dh;
+        double k = dk;
+        double hk = h * k;
+        double bvn = 0.0;
+        std::vector<double> w, x;
+
+        if (std::abs(r) < 0.3) {
```
dschmitz89: Same as above here. Can we separate out the construction of the points into a separate function?

dschmitz89 (commented):
Thanks @fbourgey . The code looks correct. I have comments regarding readability/maintainability.

Could we rename the function for example? `bvnu` has the same charm as the ancient Fortran routines in `amos`. I think it was just taken over but why not use the chance to rename them to something more descriptive?

fbourgey:
@dschmitz89 thanks for the review! I have used a more explicit name and tried to factor out some of the functions. Let me know what you think.

dschmitz89:
> @dschmitz89 thanks for the review! I have used a more explicit name and tried to factor out some of the functions. Let me know what you think.

Thanks @fbourgey , imo the changes brought a big maintainability win. I will give this my approval. Are there any extra tests for the 2D case in SciPy itself?

fbourgey:
> > @dschmitz89 thanks for the review! I have used a more explicit name and tried to factor out some of the functions. Let me know what you think.
> 
> Thanks @fbourgey , imo the changes brought a big maintainability win. I will give this my approval. Are there any extra tests for the 2D case in SciPy itself?

I found a few in `test_multivariate.py`. For example:

https://github.com/scipy/scipy/blob/main/scipy/stats/tests/test_multivariate.py#L887
https://github.com/scipy/scipy/blob/main/scipy/stats/tests/test_multivariate.py#L1016

fbourgey:
@steppi @mdhaber what do you think of those changes?

steppi:
> @steppi @mdhaber what do you think of those changes?

It looks good from an algorithmic perspective but it's not there yet since it still relies on internal `std::vector` allocations that won't work with CUDA. I finally have https://github.com/scipy/scipy/pull/24894 working now, so my recommendation is to pick one of your PRs and try to update it to work with SciPy by following that approach. My hope is that `xsf` will eventually only contain simple numerical kernels that work on both CPU and GPU with CPU specific code living in SciPy and GPU specific code living in CuPy.

Feel free to email me directly if you have any questions or need help with anything.

fbourgey:
> > @steppi @mdhaber what do you think of those changes?
> 
> It looks good from an algorithmic perspective but it's not there yet since it still relies on internal `std::vector` allocations that won't work with CUDA. I finally have [scipy/scipy#24894](https://github.com/scipy/scipy/pull/24894) working now, so my recommendation is to pick one of your PRs and try to update it to work with SciPy by following that approach. My hope is that `xsf` will eventually only contain simple numerical kernels that work on both CPU and GPU with CPU specific code living in SciPy and GPU specific code living in CuPy.
> 
> Feel free to email me directly if you have any questions or need help with anything.

@steppi @dschmitz89 I've tried to remove any `std::vector` and make the code CUDA compatible. Let me know your thoughts.

steppi:
Oh very cool. I guess I didn't look closely enough and missed that the internal allocation was completely unnecessary. This won't even need the pattern from my recent `poisson_binom` SciPy PR. I can take a closer look after I work through my backlog a bit; or @dschmitz89 feel free to review and merge this one if it looks good to you.

dschmitz89:
@fbourgey could you add a handful of tests against the current Scipy implementation? I am happy to merge after that. The tolerancecdoes not have to be super tight. Then I am happy to merge.

fbourgey:
> @fbourgey could you add a handful of tests against the current Scipy implementation? I am happy to merge after that. The tolerancecdoes not have to be super tight. Then I am happy to merge.

@dschmitz89 I have added the reference values from SciPy. Let me know if that works for you.

dschmitz89 (approved):

dschmitz89:
Thanks @fbourgey .

PR #116, opened 2026-03-27
ENH: template-ify `lambertw` with overloads for `double`, `float`, and `complex<float>`

kandersolar:
This PR adds overloads for `lambertw` for `double`, `float`, and `complex<float>` inputs.  The implementation follows @steppi's suggested approach in this forum thread: https://discuss.scientific-python.org/t/scipy-special-faster-real-only-versions-of-special-functions/2271/4

Please note that this is my first time dealing with modern C++.  Please review the code assuming it was written by someone that figured it out as they went :)

I can say at least that it compiles and passes `pixi run tests`, which I take to mean that I did not break the existing `complex<double>` path.  However, I'll need some guidance with creating tests for the new dtypes.  I see there are parquet files with test values, but I'm lost beyond that.

### DIFF
--- a/include/xsf/evalpoly.h
+++ b/include/xsf/evalpoly.h
@@ -20,28 +20,41 @@
 #pragma once
 
 #include "config.h"
+#include "cephes/polevl.h"
 
 namespace xsf {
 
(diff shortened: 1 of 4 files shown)


kandersolar:
Closing, redundant after #119

PR #123, opened 2026-04-13
ENH: calculation of `trigamma` and `polygamma` functions for complex inputs

LorenzoPeri17:
Closes issue #120 and [#7410 in scipy](https://github.com/scipy/scipy/issues/7410).

I have started a PR to get the ball rolling. I have implemented the tests myself calculating raw input-output pairs via mpmath because I could not quite wrap my head the table generation system of `xsref`. Happy to help in this effort with some guidance :)

I would also be happy to help with the plumbing to get `trigamma` and `poligamma` plumbed into `scipy.special` if there is a need!

I have taken the liberty of making a bespoke `trigamma` function because it is the one that comes up regularly in my work (and it's the original reason why I went down the polygamma rabbit hole), so I could use the special handling of the roots ;)


steppi:
Thanks @LorenzoPeri17! Don't worry about `xsref`, that's not the process contributors have been using to add tests. You can instead add tests to the `xsf_tests` folder https://github.com/scipy/xsf/tree/main/tests/xsf_tests. `scipy_special_tests` was created through a process of automatically generating tests for the existing `scipy.special` test cases from the SciPy repo, and all the infra around it really only makes sense for large scale automated test generation.

fbourgey:
Thanks @LorenzoPeri17 

CI is complaining because your code is not formatted. 

You can run `clang-format -i my_file.cpp` to format using `clang-format`

LorenzoPeri17:
Thanks for the comments! I just realized that I ran `clang-format` on the headers only, not the tests. Will run it again!

And thanks for the clarification @steppi about `xsref`! Is it then fine if I move the tests as the are (with the precomputed values from `mpmath` as tables) and move them to the `xsf_tests` folder?

PR #124, opened 2026-04-13
ENH: Add CDF and SF of the Wilcoxon distribution

fbourgey:
### Reference issue

- Implement the CDF and SF of the Wilcoxon signed-rank distribution.
- Port the exact distribution logic from [scipy.stats._wilcoxon.py](https://github.com/scipy/scipy/blob/a4463edad21e858fb7556b4fd121463da8da2e33/scipy/stats/_wilcoxon.py#L13)
- Precompute a CDF table from the PMF so repeated CDF/SF queries for a given `n` are efficient.

### DIFF
--- a/include/xsf/stats.h
+++ b/include/xsf/stats.h
@@ -502,6 +502,85 @@ inline float tukeylambdacdf(float x, double lmbda) {
     return tukeylambdacdf(static_cast<double>(x), static_cast<double>(lmbda));
 }
 
+inline std::vector<double> wilcoxon_pmf(int n) {
+    // PMF of the Wilcoxon signed-rank statistic.
+    // The returned vector has length n*(n+1)/2 + 1, and the k-th element contains
+    // the probability of observing a test statistic equal to k for n pairs of
+    // observations.
+    std::vector<double> pmf = {1.0};
+    for (int k = 1; k < n + 1; k++) {
+        std::vector<double> tmp(k * (k + 1) / 2 + 1, 0.0);
+        int m = static_cast<int>(pmf.size());
+        for (int i = 0; i < m; i++) {
+            tmp[i] += 0.5 * pmf[i];
+            tmp[i + k] += 0.5 * pmf[i];
+        }
+        pmf = std::move(tmp);
+    }
+    return pmf;
+}
+
+inline std::vector<double> wilcoxon_cdf_table(const std::vector<double> &pmf) {
+    // Cumulative probabilities for CDF/SF queries.
+    std::vector<double> cdf_table(pmf.size());
+    double sum = 0.0;
+    for (int i = 0; i < static_cast<int>(pmf.size()); ++i) {
+        sum += pmf[i];
+        cdf_table[i] = sum;
+    }
+    return cdf_table;
+}
+
+inline double wilcoxon_cdf1(int k, const std::vector<double> &cdf_table, int n) {
+    const int max_k = n * (n + 1) / 2;
+    if (k < 0) {
+        return 0.0;
+    }
+    if (k >= max_k) {
+        return 1.0;
+    }
+    return cdf_table[k];
+}
+
+inline double wilcoxon_sf1(int k, const std::vector<double> &cdf_table, int n) {
+    const int max_k = n * (n + 1) / 2;
+    if (k < 0) {
+        return 1.0;
+    }
+    if (k > max_k) {
+        return 0.0;
+    }
+    return 1.0 - (k > 0 ? cdf_table[k - 1] : 0.0);
+}
+
+inline double wilcoxon_cdf(int k, const std::vector<double> &cdf_table, int n) {
+    // CDF of the Wilcoxon signed-rank statistic for integer k and n pairs of observations.
+    const double mn = n * (n + 1) / 4.0;
+    return (k <= mn) ? wilcoxon_cdf1(k, cdf_table, n) : 1.0 - wilcoxon_sf1(k + 1, cdf_table, n);
+}
+
+inline double wilcoxon_cdf(double k, const std::vector<double> &cdf_table, int n) {
+    // CDF of the Wilcoxon signed-rank statistic for real k and n pairs of observations.
+    if (std::isnan(k) || n < 0) {
+        return std::numeric_limits<double>::quiet_NaN();
+    }
+    return wilcoxon_cdf(static_cast<int>(k), cdf_table, n);
+}
+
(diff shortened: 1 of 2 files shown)


include/xsf/stats.h
```diff
@@ -300,18 +300,92 @@ inline float tukeylambdacdf(float x, double lmbda) {
...
+        std::vector<double> tmp(k * (k + 1) / 2 + 1, 0.0);
+        int m = static_cast<int>(pmf.size());
+        for (int i = 0; i < m; i++) {
+            tmp[i] += 0.5 * pmf[i];
+            tmp[i + k] += 0.5 * pmf[i];
+        }
+        pmf = std::move(tmp);
+    }
+    return pmf;
+}
```
fbourgey: From https://github.com/scipy/scipy/blob/main/scipy/stats/_hypotests.py#L697-L711

include/xsf/stats.h
```diff
@@ -300,6 +300,85 @@ inline float tukeylambdacdf(float x, double lmbda) {
...
+    return (k <= mn) ? wilcoxon_cdf1(k, cdf_table, n) : 1.0 - wilcoxon_sf1(k + 1, cdf_table, n);
+}
+
+inline double wilcoxon_cdf(double k, const std::vector<double> &cdf_table, int n) {
+    // CDF of the Wilcoxon signed-rank statistic for real k and n pairs of observations.
+    if (std::isnan(k) || n < 0) {
+        return std::numeric_limits<double>::quiet_NaN();
+    }
+    return wilcoxon_cdf(static_cast<int>(k), cdf_table, n);
+}
```
dschmitz89: Maybe I am blind but I do not understand how this function would be called inside SciPy. Should we first generate the CDF table and then call this method? That would be highly unregular pattern for our special functions. I would expect there to be a simple `wilcoxon_cdf` function. Is this difficult here for performance reasons?

steppi: @dschmitz89, I've worked out a pattern where the results of computations which are based on only some of a ufunc's args can be reused across iterations over the other args. See for instance this stateful kernel for for the poisson binomial pmf in SciPy

https://github.com/scipy/scipy/blob/851124cf2893187a07ade72d757938cc145b79fd/scipy/special/_gufuncs.cpp#L72-L107

```C++
/* gufunc kernels which use internal caches.
 *
 * Such caches live only during the course of a single call to the ufunc
 * in order to eliminate redundant calculations. This does not concern
 * user-facing caching across ufunc calls.
 *
 * In order for such caches to serve their purpose, the iteration order over the
 * course of the ufunc loops must be chosen so that elements (or core slices) of
 * the inputs that go into the cached computation vary most slowly and elements
 * of the inputs for which the cached result can be reused vary most quickly.
 * This can be accomplished by wrapping the gufunc with `_with_cache_optimization`
 * from ./_ufunc_tools.py. See the docstring of `_with_cache_optimization` for
 * more information.
 */


template <typename T_1d>
struct _poisson_binom_pmf_kernel {
    using T = typename T_1d::value_type;


    std::vector<T> dist;
    T* last_p_ptr = nullptr;
    T operator()(long long int k, T_1d p) {
        if (!last_p_ptr) {
            dist.resize(p.extent(0) + 1);
        }
        if (p.data_handle() != last_p_ptr) {
            /* If p is stepped through in the outer loops and k in the inner loops,
             * this will yield a cache hit for each inner iteration. The cache is
             * overwritten unconditionally whenever this steps to a new slice of p. */
            xsf::poisson_binom_pmf_all(p, _as_mdspan(dist));
            last_p_ptr = p.data_handle();
        }


        return xsf::take_from_pmf(_as_mdspan(dist), k);
    }
};
```

what @fbourgey has here is not quite right though. To ensure that these `xsf` kernels will also work in CuPy, they should be using template arguments meant to be instantiated with `mdspans`, rather than using `std::vector`.  The kernels for generating the pmf and cdf tables should follow the same pattern used for the corresponding Poisson Binom kernels

https://github.com/scipy/xsf/blob/f7b85f505a94fd398024478d2e747b01dd2d7a6d/include/xsf/stats.h#L538-L578

Also @fbourgey, you should be using `xsf::take_from_discrete_cdf` and `xsf::take_from_pmf`

https://github.com/scipy/xsf/blob/f7b85f505a94fd398024478d2e747b01dd2d7a6d/include/xsf/stats.h#L580-L601

not writing distribution specific functions to do this work.

fbourgey: Thanks all. I need to update this PR following what @steppi implemented for `poisson_binom_pmf`. Will try to do it asap. This PR was submitted before we added this pattern. Changing to draft for now.

include/xsf/stats.h
```diff
@@ -300,6 +300,85 @@ inline float tukeylambdacdf(float x, double lmbda) {
...
 }
 
+inline std::vector<double> wilcoxon_pmf(int n) {
+    // PMF of the Wilcoxon signed-rank statistic.
+    // The returned vector has length n*(n+1)/2 + 1, and the k-th element contains
+    // the probability of observing a test statistic equal to k for n pairs of
+    // observations.
+    std::vector<double> pmf = {1.0};
+    for (int k = 1; k < n + 1; k++) {
+        std::vector<double> tmp(k * (k + 1) / 2 + 1, 0.0);
```
dschmitz89: I remember that @steppi  mentioned in another review that we should use `mdspan` instead of `std::vector`.

fbourgey:
I think this is ready for review.

dschmitz89:
I will happily merge after the poisson binom changes are accepted in this form and merged.

include/xsf/stats.h
```diff
@@ -600,6 +600,123 @@ XSF_HOST_DEVICE inline typename InputMat::value_type take_from_discrete_cdf(Inpu
...
+
+} // namespace detail
+
+template <typename OutputMat>
+XSF_HOST_DEVICE inline void wilcoxon_pmf_all(int n, OutputMat res) {
+    /* Fill `res` with the full PMF table for a Wilcoxon signed-rank statistic.
+     *
+     * res should be an mdspan view of a 1d array of length
+     * n * (n + 1) / 2 + 1. It is up to the caller to pass valid values
+     * for n and res.
```
dschmitz89:
```suggestion
     * res should be an mdspan view of a 1d array of length
     n * (n + 1) / 2 + 1. It is up to the caller to pass valid values
     * for n and res.
```

Typo?

fbourgey: Updated.

Issue #127, opened 2026-04-19
MAINT: add `gdtrix`, `gdtria`, `nrdtrimn` and `nrdtrisd`


dschmitz89:
Over at SciPy, I migrated the former cdflib functions `gdtrix`, `gdtria`, `nrdtrimn` and `nrdtrisd` to use `xsf` under the hood. The current state is that they awkwardly live in scipy outside of scipy's new ufunc machinery.

The better solution is that they become standard functions in `xsf` itself.

The functions are very simple, basically wrappers around `gammaincinv` and `ndtri`. Their implementations can be found here:
* [`nrdtrimn`](https://github.com/scipy/scipy/blob/87a6d10b775ed22808df8f35d92ee1b16c647610/scipy/special/xsf_wrappers.cpp#L647)
* [`nrdtrisd`](https://github.com/scipy/scipy/blob/87a6d10b775ed22808df8f35d92ee1b16c647610/scipy/special/xsf_wrappers.cpp#L660)
* [`gdtrix`](https://github.com/scipy/scipy/blob/87a6d10b775ed22808df8f35d92ee1b16c647610/scipy/special/xsf_wrappers.cpp#L608)
* [`gdtria`](https://github.com/scipy/scipy/blob/87a6d10b775ed22808df8f35d92ee1b16c647610/scipy/special/xsf_wrappers.cpp#L595)

CodeHermit17:
I want to work on this. After reviewing the codebase, I see `gdtrib` is already implemented in `cdflib.h`, so I plan to add `gdtrix`, `gdtria`, `nrdtrimn`, and `nrdtrisd` there as well, following the same pattern. Let me know if that's the right approach before I start.

fbourgey:
With #133 and #160 in, let's close this. @dschmitz89 I will follow up with a SciPy PR reflecting those changes.

PR #128, opened 2026-04-19
ENH: add `nrdtrimn`

dschmitz89:
Towards #127 

This PR adds the function `nrdtrimn` abd is also very much educational to familiarize myself with the xsf repo. About this: as a non seasoned C++ coder I found the developer experience actually pretty great. Compiling xsf is simpler than SciPy thanks to the Pixi setup (hat tip @lucascolley @steppi ). One nit is that I currently do not understand how to run only one test, maybe add this to the README?

I added only one test with real values as `ndtri` is likely tested a lot harder already.

Note that the edge case handling was simply copied over from SciPy which mostly preserved backwards compatibility to the ancient cdflib. I am not sure if `ndtri` is currently GPU capable and GPU tested so I did not add the appropriate macros.

### DIFF
--- a/include/xsf/ndtri_reparametrizations.h
+++ b/include/xsf/ndtri_reparametrizations.h
@@ -1,0 +1,35 @@
+#pragma once
+
+#include <cmath>
+#include <limits>
+
+#include "xsf/cephes/ndtri.h"
+
+namespace xsf {
+namespace detail {
+
+    // Returns the mean of a normal distribution with standard deviation std
+    // such that the CDF at x equals p.
+    inline double nrdtrimn(double p, double std, double x) {
+        if (std::isnan(std) || std <= 0) {
+            return std::numeric_limits<double>::quiet_NaN();
+        }
+        if (std::isnan(p) || p <= 0 || p >= 1) {
+            return std::numeric_limits<double>::quiet_NaN();
+        }
+        if (std::isnan(x)) {
+            return std::numeric_limits<double>::quiet_NaN();
+        }
+        return x - std * cephes::ndtri(p);
+    }
+
+} // namespace detail
+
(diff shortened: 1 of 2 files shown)


steppi:
Thanks @dschmitz89! Currently there is no way to run only one test with a pixi command. What I do currently is resort to invoking `ctest` directly with the `-R` flag to pick tests by regex.

dschmitz89:
I ended up using `cmake --build build --target [testfile]` and then simply executing the test.

About this PR: Let's use this PR as a testbed for the rest of #127 ? Once we agree on where to put them and how strong they need to be tested, adding the rest should be straight forward. Could leave this to new contributors (cc @CodeHermit17) ) for example.

CodeHermit17:
Hi @dschmitz89, thanks for the tag! I've implemented `gdtrix`, `gdtria` and `nrdtrisd` with tests and a few domain validations. I'd be happy to contribute these.
Would you like me to open a separate PR targeting your branch, or should I just open a new PR against main that references this one?

dschmitz89:
> Hi @dschmitz89, thanks for the tag! I've implemented `gdtrix`, `gdtria` and `nrdtrisd` with tests and a few domain validations. I'd be happy to contribute these.
> Would you like me to open a separate PR targeting your branch, or should I just open a new PR against main that references this one?
> 

Hi @CodeHermit17 I suggest to open one PR per function at a time, that reduces the review load for xsf maintainers.

fbourgey:
Should they be added to `/xsf/stats.h`? 

cc @steppi

dschmitz89:
> Should they be added to `/xsf/stats.h`?
> 
> cc @steppi

That seems sensible. Thoughts @steppi ?

steppi:
Sorry, missed the ping due to issues with my email filters. Yes,`xsf/stats.h` sounds good to me.

PR #130, opened 2026-04-19
ENH: Add inverse digamma function

### LINKED ISSUE #129: ENH: Inverse digamma function
In https://github.com/scipy/scipy/issues/17507 a user asked for the inverse digamma function. As its purpose is mostly for maximum likelihood fitting of real continuous distributions it should be fine to restrict this to real inputs first and ignore complex ones.

The SciPy issue contains a very good [reference](https://tminka.github.io/papers/dirichlet/minka-dirichlet.pdf) which lists in appendix C a simple Newtonian root finder with suitable initial guesses. The algorithm looks actually quite simple to implement. Does xsf already contain helper functions for such derivative based root finding/fix point problems?

dschmitz89:
Closes #129 

Full disclosure: Most of the actual code was generated by a LLM. I pointed it to the reference and fine tuned the convergence criterion and the edge cases.

We will likely need to add a lot more tests, I will look into that in the coming weeks.


include/xsf/digamma_inv.h
```diff
@@ -0,0 +1,66 @@
...
+        //   T. Minka, "Estimating a Dirichlet distribution", 2000. Appendix C
+        //   https://tminka.github.io/papers/dirichlet/minka-dirichlet.pdf
+        if (std::isnan(y)) {
+            return std::numeric_limits<double>::quiet_NaN();
+        }
+        if (std::isinf(y)) {
+            // digamma(x) -> +inf as x -> +inf, and digamma(x) -> -inf as x -> 0+.
+            return y > 0 ? std::numeric_limits<double>::infinity() : 0.0;
+        }
+
```
fbourgey: should we also add a small guard?

```cpp
if (y > std::log(std::numeric_limits<double>::max())) {
      }
```

since for finite large `y` (ie above log(DBL_MAX), `std::exp(y)` can overflow before Newton starts (the Newton step evaluates roughly `(psi(inf) - y) / zeta(2,inf)` and would produce a NaN.

A regression test for `y=710.0~log(DBL_MAX)` would also be needed.

dschmitz89:
This PR is ready from my side.

tests/xsf_tests/test_digamma_inv.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <xsf/cephes/psi.h>
+#include <xsf/digamma_inv.h>
+
+#include <cmath>
+#include <limits>
+
+TEST_CASE("digamma_inv round-trip", "[digamma_inv][xsf_tests]") {
+    const double rtol = 100 * std::numeric_limits<double>::epsilon();
+
+    const std::vector<double> xs = {1e-30, 1e-10, 1e-5, 0.1, 1.0, 2.0, 5.0, 10.0, 100.0};
```
fbourgey: Maybe we could use a log-spaced grid with [`linspace`](https://github.com/scipy/xsf/blob/main/tests/testing_utils.h#L159) to cover more points.

Adding a `logspace` function in `tests/testing_utils.h` would be useful too.

dschmitz89: Looks like we both pushed this at the same time. Is what I did here what you had in mind?

fbourgey: Yes exactly thanks. I am going to close https://github.com/scipy/xsf/pull/136.

include/xsf/digamma_inv.h
```diff
@@ -0,0 +1,70 @@
...
+    // Refine an initial guess x using Newton-Raphson iteration:
+    //   x_new = x_old - (psi(x_old) - y) / psi'(x_old)
+    // where psi'(x) = zeta(2, x) (the trigamma function).
+    // x is passed by value; the refined result is returned.
+    inline double digamma_inv_newton_raphson(double x, double y) {
+        const int max_iterations = 20; // avoid infinite loops in pathological cases
+        for (int i = 0; i < max_iterations; ++i) {
+            double psi_x = cephes::psi(x);
+            double psi1_x = cephes::zeta(2, x); // trigamma: psi'(x) = zeta(2, x)
```
fbourgey:
```suggestion
            double trigamma_x = cephes::zeta(2, x); // trigamma: psi'(x) = zeta(2, x)
```

fbourgey (commented):
LGTM! 

However, would it make sense to add `XSF_HOST_DEVICE`  to allow compilation for both CPU and CUDA device code? cc @steppi to confirm.

fbourgey:
For future reference, this PR implements 
Screenshot 2026-05-05 at 4 34 32 PM

Reference: [T. Minka, "Estimating a Dirichlet distribution", 2000. Appendix C](https://tminka.github.io/papers/dirichlet/minka-dirichlet.pdf)

include/xsf/digamma_inv.h
```diff
@@ -0,0 +1,70 @@
+#pragma once
+
+#include <cmath>
+#include <limits>
+
+#include "xsf/cephes/psi.h"
+#include "xsf/cephes/zeta.h"
+
```
steppi:
```suggestion

#include "xsf/config.h"
#include "xsf/cephes/psi.h"
#include "xsf/cephes/zeta.h"

```

Instead of including these stdlib headers directly, the pattern in `xsf` is to include [`config.h`](https://github.com/scipy/xsf/blob/main/include/xsf/config.h) which includes the headers. `config.h` uses conditional compilation to check if the header is being compiled in a C++ or CUDA program, and does the moral equivalent of treating `std` as an alias for `cuda::std`.

Without including `config.h` you won't be able to use the XSF_HOST_DEVICE macro that @fbourgey suggested adding.

dschmitz89:
Reviews were addressed, let me know if anything else is missing. I did not test GPU support though.

steppi (approved):
Thanks @dschmitz89. Don't worry about GPU tests. The main point of those is to spot CuPy regressions so they don't show up for the first time when making a PR to update `xsf` in CuPy. We just need to make sure we add them before we eventually submit a CuPy PR for inverse digamma.

I'll merge this one now. For the immediate future, also feel free to merge on your own if you get a green checkmark and the go-ahead from @fbourgey.

PR #131, opened 2026-04-21
Fix j0 for large arguments

dschmitz89:
Motivated by https://github.com/scipy/scipy/issues/22705 over at SciPy.

The fix is directly taken from @pearu 's suggestion: rewriting 
`sqrt(2 / (pi * x)) * cos(x - pi/4) = sqrt(1 / (pi * x)) * (cos(x) + sin(x))`.

The same fix can be applied to `y0` in a followup.

### DIFF
--- a/include/xsf/cephes/j0.h
+++ b/include/xsf/cephes/j0.h
@@ -160,7 +160,7 @@ namespace cephes {
     } // namespace detail
 
     XSF_HOST_DEVICE inline double j0(double x) {
-        double w, z, p, q, xn;
+        double w, z, p, q;
 
         if (x < 0) {
             x = -x;
@@ -181,9 +181,8 @@ namespace cephes {
         q = 25.0 / (x * x);
         p = polevl(q, detail::j0_PP, 6) / polevl(q, detail::j0_PQ, 6);
         q = polevl(q, detail::j0_QP, 7) / p1evl(q, detail::j0_QQ, 7);
-        xn = x - M_PI_4;
-        p = p * std::cos(xn) - w * q * std::sin(xn);
-        return (p * detail::SQRT2OPI / std::sqrt(x));
+        p = (p + w * q) * std::cos(x) + (p - w * q) * std::sin(x);
+        return (p * detail::SQRT1OPI / std::sqrt(x));
     }
 
     /*                                                     y0() 2  */
--- a/include/xsf/cephes/j0.h
+++ b/include/xsf/cephes/j0.h
@@ -181,6 +181,11 @@ namespace cephes {
         q = 25.0 / (x * x);
         p = polevl(q, detail::j0_PP, 6) / polevl(q, detail::j0_PQ, 6);
         q = polevl(q, detail::j0_QP, 7) / p1evl(q, detail::j0_QQ, 7);
+        if (x < 10.0) {
+            double xn = x - M_PI_4;
+            p = p * std::cos(xn) - w * q * std::sin(xn);
+            return (p * detail::SQRT2OPI / std::sqrt(x));
+        }
         p = (p + w * q) * std::cos(x) + (p - w * q) * std::sin(x);
         return (p * detail::SQRT1OPI / std::sqrt(x));
     }
--- a/tests/xsf_tests/test_j0_right_tail.cpp
+++ b/tests/xsf_tests/test_j0_right_tail.cpp
@@ -1,0 +1,15 @@
+#include "../testing_utils.h"
+#include <xsf/cephes/j0.h>
+
+TEST_CASE("j0 right tail gh-large-input", "[j0][xsf_tests]") {
+    // Reference value computed with mpmath: mp.besselj(0, 1e15)
+    const double x = 1e15;
+    const double ref = 6.156638646885021e-09;
+    const double rtol = 1e-13;
+
+    const double w = xsf::cephes::j0(x);
+    const double rel_error = xsf::extended_relative_error(w, ref);
+
+    CAPTURE(x, w, ref, rtol, rel_error);
+    REQUIRE(rel_error <= rtol);
+}
--- a/tests/xsf_tests/test_j0_right_tail.cpp
+++ b/tests/xsf_tests/test_j0_right_tail.cpp
@@ -2,14 +2,16 @@
 #include <xsf/cephes/j0.h>
 
 TEST_CASE("j0 right tail gh-large-input", "[j0][xsf_tests]") {
-    // Reference value computed with mpmath: mp.besselj(0, 1e15)
-    const double x = 1e15;
-    const double ref = 6.156638646885021e-09;
-    const double rtol = 1e-13;
-
+    using test_case = std::tuple<double, double>;
+    // Reference values computed with mpmath with 1000 digts of precision
+    auto [x, ref] = GENERATE(
+        test_case{1e15, 6.156638646885021e-09}, test_case{1e12, 1.0167125050040682e-07},
+        test_case{1e13, 1.1926484739665653e-07}, test_case{1e8, 3.206029534041208e-05},
+        test_case{1e20, 6.698009040703424e-12}
+    );
     const double w = xsf::cephes::j0(x);
     const double rel_error = xsf::extended_relative_error(w, ref);
 
-    CAPTURE(x, w, ref, rtol, rel_error);
-    REQUIRE(rel_error <= rtol);
+    CAPTURE(x, w, ref, rel_error);
+    REQUIRE(rel_error <= 1e-15);
 }
--- a/tests/xsf_tests/test_j0_right_tail.cpp
+++ b/tests/xsf_tests/test_j0_right_tail.cpp
@@ -3,15 +3,27 @@
 
 TEST_CASE("j0 right tail gh-large-input", "[j0][xsf_tests]") {
     using test_case = std::tuple<double, double>;
-    // Reference values computed with mpmath with 1000 digts of precision
-    auto [x, ref] = GENERATE(
-        test_case{1e15, 6.156638646885021e-09}, test_case{1e12, 1.0167125050040682e-07},
-        test_case{1e13, 1.1926484739665653e-07}, test_case{1e8, 3.206029534041208e-05},
-        test_case{1e20, 6.698009040703424e-12}
+    // Reference values computed with mpmath with 1000 digits of precision:
+    //
+    // import math
+    // import mpmath as mp
+    // mp.mp.dps = 1000
+    // xs = [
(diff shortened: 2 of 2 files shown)


fbourgey:
Screenshot 2026-04-21 at 4 44 13 PM

The tolerance might need to be lowered for this to pass

dschmitz89:
> Screenshot 2026-04-21 at 4 44 13 PM
> The tolerance might need to be lowered for this to pass

Thanks for taking a look. Where do those reference values actually come from? I would like to check first if the new value is indeed worse. `cyl_bessel_j0` simply dispatches to `j0` from what I understand.

fbourgey:
I think they come from mpmath https://github.com/scipy/xsref

cc @steppi to confirm

steppi:
Thanks for working on this @dschmitz89 and on the other PRs. Yes, the reference implementation is based on `mpmath`. 

https://github.com/scipy/xsref/blob/9661be503b6d8822966c34dab9549097c0b84ca6/src/xsref/_reference/_functions.py#L462-L466

I was planning to use some time last week to improve reference table side of things so tolerances can be updated more easily, but ended up being out sick most of the week.  In this case, I think we can just branch on the size of $|x|$ though to decide which formula to use. After taking sick leave, I'll need to catch up on some other things before I can justify using leave to spend unfunded time on `xsf`, but I hope to get to it sooner than later, since it leads to a lot of annoyances that the table based tests are in an unfinished state.

include/xsf/cephes/j0.h
```diff
@@ -181,9 +181,8 @@ namespace cephes {
         q = 25.0 / (x * x);
         p = polevl(q, detail::j0_PP, 6) / polevl(q, detail::j0_PQ, 6);
         q = polevl(q, detail::j0_QP, 7) / p1evl(q, detail::j0_QQ, 7);
-        xn = x - M_PI_4;
-        p = p * std::cos(xn) - w * q * std::sin(xn);
-        return (p * detail::SQRT2OPI / std::sqrt(x));
+        p = (p + w * q) * std::cos(x) + (p - w * q) * std::sin(x);
+        return (p * detail::SQRT1OPI / std::sqrt(x));
```
fbourgey, line 190:
```suggestion
        if (x < 10.0) {
            double xn = x - M_PI_4;
            p = p * std::cos(xn) - w * q * std::sin(xn);
            return (p * detail::SQRT2OPI / std::sqrt(x));
        }
        p = (p + w * q) * std::cos(x) + (p - w * q) * std::sin(x);
        return (p * detail::SQRT1OPI / std::sqrt(x));
```

Following @steppi's [comment](https://github.com/scipy/xsf/pull/131#issuecomment-4297487350), a cutoff at `x < 10` seems to do the trick.

tests/xsf_tests/test_j0_right_tail.cpp
```diff
@@ -0,0 +1,17 @@
+#include "../testing_utils.h"
+#include <xsf/cephes/j0.h>
+
+TEST_CASE("j0 right tail gh-large-input", "[j0][xsf_tests]") {
+    using test_case = std::tuple<double, double>;
+    // Reference values computed with mpmath with 1000 digts of precision
```
fbourgey:
```suggestion
    // Reference values computed with mpmath with 1000 digits of precision:
    //
    // import math
    // import mpmath as mp
    // mp.mp.dps = 1000
    // xs = [
    //     math.nextafter(10.0, 0.0), 10.0, math.nextafter(10.0, math.inf),
    //     100.0, 1e4, 1e8, 1e12, 1e13, 1e15, 1e20,
    // ]
    // for x in xs:
    //     print(x, mp.nstr(mp.besselj(0, x), 18))
```

I suggest adding the mpmath code

tests/xsf_tests/test_j0_right_tail.cpp
```diff
@@ -0,0 +1,17 @@
...
+    auto [x, ref] = GENERATE(
+        test_case{1e15, 6.156638646885021e-09}, test_case{1e12, 1.0167125050040682e-07},
+        test_case{1e13, 1.1926484739665653e-07}, test_case{1e8, 3.206029534041208e-05},
+        test_case{1e20, 6.698009040703424e-12}
+    );
+    const double w = xsf::cephes::j0(x);
+    const double rel_error = xsf::extended_relative_error(w, ref);
+
+    CAPTURE(x, w, ref, rel_error);
+    REQUIRE(rel_error <= 1e-15);
```
fbourgey:
```suggestion
    auto [x, ref, rtol] = GENERATE(
        test_case{9.999999999999998, -0.24593576445134826, 5e-15}, test_case{10.0, -0.24593576445134834, 5e-15},
        test_case{10.000000000000002, -0.24593576445134841, 5e-15}, test_case{100.0, 0.019985850304223122, 5e-15},
        test_case{1e4, -0.0070961603533888015, 5e-15}, test_case{1e8, 3.206029534041208e-05, 1e-15},
        test_case{1e12, 1.0167125050040682e-07, 1e-15}, test_case{1e13, 1.1926484739665653e-07, 1e-15},
        test_case{1e15, 6.156638646885021e-09, 1e-15}, test_case{1e20, 6.698009040703424e-12, 2e-15}
    );
    const double w = xsf::cephes::j0(x);
    const double rel_error = xsf::extended_relative_error(w, ref);

    CAPTURE(x, w, ref, rtol, rel_error);
    REQUIRE(rel_error <= rtol);
```

Strengthening the test and adding values before, at, and after the cutoff. Also, small nit, it seems more natural to have increasing values for `x`.

include/xsf/cephes/j0.h
dschmitz89: That will likely work but did you do a benchmark against a mpmath reference to get the cutoff value?

dschmitz89:
Thanks for the review and the computation of the mpmath references @fbourgey! I will commit your suggestions when I understand how exactly you came up with the cutoff value of 10.

fbourgey:
> Thanks for the review and the computation of the mpmath references @fbourgey! I will commit your suggestions when I understand how exactly you came up with the cutoff value of 10.

Codex helped me write a script reproducing the old and new `j0` tail formulas. It compares both against high-precision `mpmath.besselj(0, x)` over ranges above `x = 5`.
It shows the old formula is slightly better on `5 < x < 10`, while the rewritten formula is much better from about `x >= 10`, so `10` seemed like a sensible cutoff.

I then tested this locally, and this seems to work robustly. Let me know your thoughts. There might be a better cutoff, though, and 10 might not be the optimal one.

<details>

```python
import math

import mpmath as mp

mp.mp.dps = 100

# Cephes j0 tail coefficients from include/xsf/cephes/j0.h
PP = [
    7.96936729297347051624e-4,
    8.28352392107440799803e-2,
    1.23953371646414299388e0,
    5.44725003058768775090e0,
    8.74716500199817011941e0,
    5.30324038235394892183e0,
    9.99999999999999997821e-1,
]
PQ = [
    9.24408810558863637013e-4,
    8.56288474354474431428e-2,
    1.25352743901058953537e0,
    5.47097740330417105182e0,
    8.76190883237069594232e0,
    5.30605288235394617618e0,
    1.00000000000000000218e0,
]
QP = [
    -1.13663838898469149931e-2,
    -1.28252718670509318512e0,
    -1.95539544257735972385e1,
    -9.32060152123768231369e1,
    -1.77681167980488050595e2,
    -1.47077505154951170175e2,
    -5.14105326766599330220e1,
    -6.05014350600728481186e0,
]
QQ = [
    6.43178256118178023184e1,
    8.56430025976980587198e2,
    3.88240183605401609683e3,
    7.24046774195652478189e3,
    5.93072701187316984827e3,
    2.06209331660327847417e3,
    2.42005740240291393179e2,
]


def polevl(x, c):
    r = 0.0
    for a in c:
        r = r * x + a
    return r


def p1evl(x, c):
    r = x + c[0]
    for a in c[1:]:
        r = r * x + a
    return r


def vals(x):
    w = 5.0 / x
    z = 25.0 / (x * x)
    p = polevl(z, PP) / polevl(z, PQ)
    q = polevl(z, QP) / p1evl(z, QQ)

    old = (
        (p * math.cos(x - math.pi / 4) - w * q * math.sin(x - math.pi / 4))
        * math.sqrt(2 / math.pi)
        / math.sqrt(x)
    )

    new = (
        ((p + w * q) * math.cos(x) + (p - w * q) * math.sin(x))
        * math.sqrt(1 / math.pi)
        / math.sqrt(x)
    )

    return old, new


def rel(y, r):
    return float(abs(mp.mpf(y) - r) / abs(r)) if r else float(abs(y))


for lo, hi, n in [
    (5, 10, 1000),
    (10, 100, 1000),
    (100, 1e3, 1000),
    (1e3, 1e4, 1000),
    (1e4, 1e5, 1000),
    (1e5, 1e6, 500),
    (1e6, 1e8, 300),
]:
    oldmax = newmax = 0.0
    oldwins = newwins = 0
    worst_old = worst_new = None

    for i in range(n):
        x = lo + (hi - lo) * (i + 0.5) / n
        r = mp.besselj(0, x)
        old, new = vals(x)
        eo = rel(old, r)
        en = rel(new, r)

        if eo < en:
            oldwins += 1
        elif en < eo:
            newwins += 1

        if eo > oldmax:
            oldmax = eo
            worst_old = x
        if en > newmax:
            newmax = en
            worst_new = x

    print(
        f"{lo:g}-{hi:g}: "
        f"oldmax {oldmax:.2e}@{worst_old:.6g} "
        f"newmax {newmax:.2e}@{worst_new:.6g} "
        f"oldwins {oldwins} newwins {newwins}"
    )
```

gives the output

```markdown
5-10: oldmax 2.01e-14@8.6525 newmax 3.20e-14@8.6575 oldwins 276 newwins 445
10-100: oldmax 5.28e-12@55.765 newmax 4.78e-14@55.765 oldwins 27 newwins 943
100-1000: oldmax 1.80e-11@341.65 newmax 3.72e-14@341.65 oldwins 8 newwins 989
1000-10000: oldmax 1.78e-10@9590.5 newmax 6.15e-14@3200.5 oldwins 0 newwins 1000
10000-100000: oldmax 4.45e-10@96355 newmax 1.96e-14@70795 oldwins 0 newwins 1000
100000-1e+06: oldmax 1.24e-08@131500 newmax 9.38e-13@131500 oldwins 0 newwins 500
1e+06-1e+08: oldmax 7.33e-07@2.7235e+07 newmax 1.93e-13@2.7235e+07 oldwins 0 newwins 300
```

</details>

dschmitz89:
Is this ready to go @fbourgey ?

fbourgey:
> Is this ready to go @fbourgey ?

LGTM! thanks @dschmitz89

PR #132, opened 2026-04-24
MAINT: remove Fortran like handling of constants

dschmitz89:
This is a first step to make AMOS speak a little less FORTRAN.


include/xsf/amos/amos.h
```diff
@@ -3696,8 +3666,8 @@ namespace amos {
             //
             // SCALE BACKWARD RECURRENCE, BRY(3) IS DEFINED BUT NEVER USED
             //
-            bry[0] = 1e3 * d1mach[0] / tol;
-            bry[1] = tol / 1e3 * d1mach[0];
+            bry[0] = THRESHOLD_MIN / tol;
+            bry[1] = tol / THRESHOLD_MIN;
```
fbourgey, line 3668: Is this intended? The old code implied 

```
bry[1] = tol / 1e3 * d1mach[0] = (tol / 1e3) * d1mach[0]
```

while the new code implies 

```
bry[1] = tol / (1e3 * d1mach[0])
```

include/xsf/amos/amos.h
```diff
@@ -3352,15 +3322,15 @@ namespace amos {
         // 12 <= E <= 60. E IS COMPUTED FROM 2**(-E)=B**(1-DIGITS(0.0_dp))=
         // TOL WHERE B IS THE BASE OF THE ARITHMETIC.
         //
-        t1 = (i1mach[13] - 1) * d1mach[4] * (std::log(10) / std::log(2));
+        t1 = (std::numeric_limits<double>::digits - 1) * LOG10_2 * (std::log(10) / std::log(2));
```
fbourgey, line 3323: Can we rewrite this as

```suggestion
        t1 = std::numeric_limits<double>::digits - 1;
```

include/xsf/amos/amos.h
```diff
@@ -1032,17 +1012,16 @@ namespace amos {
         double aa, acz, aez, ak, arg, arm, atol, az, bb, bk, dfnu;
         double dnu2, fdn, rtr1, s, sgn, sqk, x, yy;
         int ib, il, inu, j, jl, k, koded, m, nn;
-        double pi = 3.14159265358979324;
-        double rpi = 0.159154943091895336; /* (1 / pi) */
```
fbourgey, line 1036: There was a typo for `rpi`. The comment should have been `/* (1 / (2*pi)) */`

include/xsf/amos/amos.h
```diff
@@ -1032,17 +1012,16 @@ namespace amos {
...
         az = std::abs(z);
         x = std::real(z);
-        arm = 1e3 * d1mach[0];
+        arm = THRESHOLD_MIN;
         rtr1 = std::sqrt(arm);
         il = (n > 2 ? 2 : n);
         dfnu = fnu + (n - il);
         // OVERFLOW TEST
-        ak1 = std::sqrt(rpi / z);
+        ak1 = std::sqrt(rtpi / z);
```
fbourgey: what about getting rid of `rtpi` and doing
```suggestion
        ak1 = std::sqrt((0.5 * M_1_PI) / z);
```
to avoid defining `rtpi` and using any confusing names?

include/xsf/amos/amos.h
JaRoSchm: Looks like this actually fixes a translation error. Here is the original Fortran code: https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zbuni.f#L54

include/xsf/amos/amos.h
```diff
@@ -92,9 +92,11 @@
  */
 #pragma once
 
+#include "../config.h"
```
dschmitz89:
```suggestion
```

include/xsf/amos/amos.h
```diff
@@ -92,9 +92,11 @@
...
 #include <cmath>
 #include <complex>
 #include <cstdlib>
+#include <limits>
```
dschmitz89:
```suggestion
#include "xsf/config.h"
```

include/xsf/amos/amos.h
```diff
@@ -1032,17 +1012,16 @@ namespace amos {
         double aa, acz, aez, ak, arg, arm, atol, az, bb, bk, dfnu;
         double dnu2, fdn, rtr1, s, sgn, sqk, x, yy;
         int ib, il, inu, j, jl, k, koded, m, nn;
-        double pi = 3.14159265358979324;
-        double rpi = 0.159154943091895336; /* (1 / pi) */
         int nz = 0;
+        double rtpi = M_1_PI / 2.0;
```
dschmitz89:
```suggestion
```

dschmitz89:
@JaRoSchm Since you are familiar with AMOS, could you have a final look here? My long term plan was to make this more maintainable by using modern idioms and removing gotos which are planned for follow ups.

include/xsf/amos/amos.h
```diff
@@ -5482,9 +5445,8 @@ namespace amos {
         std::complex<double> zeta1[2] = {0.0};
         std::complex<double> zeta2[2] = {0.0};
         std::complex<double> cy[2] = {0.0};
-        double bry[3] = {1e3 * d1mach[0] / tol, tol / 1e3 * d1mach[0], d1mach[1]};
+        double bry[3] = {THRESHOLD_MIN / tol, tol / THRESHOLD_MIN, std::numeric_limits<double>::max()};
```
JaRoSchm, line 5448: This fixes a translation bug, see https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zunk1.f#L46.

include/xsf/amos/amos.h
```diff
@@ -5876,7 +5836,9 @@ namespace amos {
         std::complex<double> asum[2] = {0.0};
         std::complex<double> bsum[2] = {0.0};
         std::complex<double> cy[2] = {0.0};
-        double bry[3] = {(1.0 + 1e3 * d1mach[0] / tol), 1.0 / (1.0 + 1e3 * d1mach[0] / tol), d1mach[1]};
+        double bry[3] = {
+            (1.0 + THRESHOLD_MIN / tol), 1.0 / (1.0 + THRESHOLD_MIN / tol), std::numeric_limits<double>::max()
+        };
```
JaRoSchm:
```suggestion
        double bry[3] = {THRESHOLD_MIN / tol, tol / THRESHOLD_MIN, std::numeric_limits<double>::max()};
```
This looked suspicious to me and is actually a translation bug, see https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zunk2.f#L60.

JaRoSchm (changes requested):
In general, looks like a nice improvement!

JaRoSchm:
> My long term plan was to make this more maintainable by using modern idioms and removing gotos which are planned for follow ups.

There's definitely potential for that in amos! In my opinion, it might be worth however to first try to fix #137 (fixing long known bugs/accuracy improvements in the Fortran implementation by the original author, which were not included in scipy before the translation) as this probably becomes harder if the versions diverge even more (and they already have quite a bit). In principle, I plan to look into this myself, but I can't say exactly when I'll get round to it yet.

include/xsf/amos/amos.h
dschmitz89: Thanks. Since this looks like the same `bry` array as your other suggestion, should we add it to the "preamble" as a `constexpr` as it is used in several places?

JaRoSchm: Sounds reasonable. If I counted correctly, there are seven places where this is defined.

dschmitz89:
@JaRoSchm Thanks. Feel free to ping me when you feel like this should be merged. You have by far the best overview of everything amos related here and we should follow the PR order you propose. I will wait with resolving the merge conflicts until then.

JaRoSchm:
@dschmitz89 I see no reason for preferring any specific order of merging the open amos-related PRs (only #92 should be merged before I can add a test to #93). My previous comment is addressed by #149.

include/xsf/amos/amos.h
```diff
@@ -1373,7 +1348,7 @@ namespace amos {
         // TEST FOR PROPER RANGE
         //
         az = std::abs(z);
-        bb = d1mach[1] * 0.5;
+        bb = std::numeric_limits<double>::max() * 0.5;
```
JaRoSchm, line 1351: This is another translation error and should have been `i1mach[8]`, see https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zbesh.f#L213.

```suggestion
        bb = std::numeric_limits<int>::max() * 0.5;
```

include/xsf/amos/amos.h
```diff
@@ -1990,7 +1961,7 @@ namespace amos {
         fn = fnu + (n - 1);
 
         aa = 0.5 / tol;
-        bb = d1mach[1] * 0.5;
+        bb = std::numeric_limits<double>::max() * 0.5;
```
JaRoSchm, line 1964: This is another translation error and should have been `i1mach[8]`, see https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zbesj.f#L192.

```suggestion
        bb = std::numeric_limits<int>::max() * 0.5;
```

JaRoSchm (changes requested):
By chance, I stumbled across two more translation errors.

dschmitz89:
@JaRoSchm Would you be willing to make a PR against my branch? I am starting to lose track of all the AMOS changes and this code is very hard to digest mentally. If not, we could still apply [this change](https://github.com/scipy/xsf/pull/132#discussion_r3224212835) and proceed in a followup.

JaRoSchm:
@dschmitz89 I think you could simply accept @fbourgey's and my suggestions here (should be possible through the GitHub UI?). Then, this is ready to merge from my pov and https://github.com/scipy/xsf/pull/132#discussion_r3224212835 could be left for a followup.

JaRoSchm (approved):

include/xsf/amos/amos.h
```diff
@@ -1373,7 +1348,7 @@ namespace amos {
         // TEST FOR PROPER RANGE
         //
         az = std::abs(z);
-        bb = d1mach[1] * 0.5;
+        bb = std::numeric_limits<int>::max() * 0.5;
```
fbourgey: From https://github.com/scipy/xsf/blob/main/include/xsf/amos/amos.h#L141, should this be?


```suggestion
        bb = std::numeric_limits<double>::max() * 0.5;
```

include/xsf/amos/amos.h
```diff
@@ -1990,7 +1961,7 @@ namespace amos {
         fn = fnu + (n - 1);
 
         aa = 0.5 / tol;
-        bb = d1mach[1] * 0.5;
+        bb = std::numeric_limits<int>::max() * 0.5;
```
fbourgey: Same here

```suggestion
        bb = std::numeric_limits<double>::max() * 0.5;
```

include/xsf/amos/amos.h
JaRoSchm: This fixes a translation bug, see https://github.com/scipy/xsf/pull/132#pullrequestreview-4318664049.

include/xsf/amos/amos.h
JaRoSchm: This fixes a translation bug, see https://github.com/scipy/xsf/pull/132#pullrequestreview-4318664049.

fbourgey:
@dschmitz89 can you please resolve the conflict? I can then take a fresh look, and get that one in if OK.

dschmitz89:
@JaRoSchm Looks like I messed something up which was fixed in main during conflict resolution. Would you have time to take a look?

JaRoSchm:
@dschmitz89 I looked at the code but cannot find what is going wrong here. In particular, I do not know why it only fails on some platforms. This would suggest that `std::numeric_limits` gives different values on the different platforms, maybe? Sadly, I won't have much time to investigate further this week.

I'd be happy if @fbourgey could also have a look, he knows C++ much better than I do.

fbourgey:
The culprits seem to be https://github.com/scipy/xsf/pull/132#pullrequestreview-4318664049. I guess we should get this PR first and then follow up with a different PR and fix the tests. This seemed out of scope here anyway. Do you agree?

JaRoSchm:
Sounds good to me!

Just a note for future reference: I think the tests that failed before you reverted the change aren't wrong. So there probably is some other problem in addition to the translation error.

PR #138, opened 2026-05-06
BUG: amos/amos.h: fix translation bugs in binu and rati

JaRoSchm:
#### Reference issue
-

Hi, I detected these errors in the same way as in https://github.com/scipy/xsf/pull/92. This time I used `besj` with `z = 70.26595687262548 - 0.7026595687262548j` and n = 50 in https://github.com/JaRoSchm/pyamos/blob/main/compare_xsf_zbessel.py for finding the bugs. This was simply incorrectly translated from FORTRAN, see https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/. Regarding testing the same question/comment from https://github.com/scipy/xsf/pull/92 applies.

#### Additional information
-

#### AI Generation Disclosure
No AI tools used

### DIFF
--- a/include/xsf/amos/amos.h
+++ b/include/xsf/amos/amos.h
@@ -2788,11 +2788,7 @@ namespace amos {
         }
         /* 100 */
         if (nw > 0) {
-            nz = -1;
-            if (nw == -2) {
-                nz = -2;
-            }
-            return nz;
+            return -1;
         }
         nw = wrsk(z, fnu, kode, nn, cy, cw, tol, elim, alim);
         if (nw < 0) {
--- a/include/xsf/amos/amos.h
+++ b/include/xsf/amos/amos.h
@@ -2666,7 +2666,7 @@ namespace amos {
         dfnu = fnu + n - 1;
         if ((az <= 2.) || (az * az * 0.25 <= (dfnu + 1.0))) {
             /* GOTO 10 */
-            nw = seri(z, fnu, kode, n, cy, tol, elim, alim);
+            nw = seri(z, fnu, kode, nn, cy, tol, elim, alim);
             inw = std::abs(nw);
             nz += inw;
             nn -= inw;
@@ -2689,7 +2689,7 @@ namespace amos {
                 //
                 // MILLER ALGORITHM NORMALIZED BY THE SERIES
                 //
-                nw = mlri(z, fnu, kode, n, cy, tol);
+                nw = mlri(z, fnu, kode, nn, cy, tol);
                 if (nw < 0) {
                     nz = -1;
                     if (nw == -2) {
@@ -2706,7 +2706,7 @@ namespace amos {
                 //
                 // ASYMPTOTIC EXPANSION FOR LARGE Z
                 //
-                nw = asyi(z, fnu, kode, n, cy, rl, tol, elim, alim);
+                nw = asyi(z, fnu, kode, nn, cy, rl, tol, elim, alim);
                 if (nw < 0) {
                     nz = -1;
                     if (nw == -2) {
@@ -2760,7 +2760,7 @@ namespace amos {
         /* 60 */
         if (az <= rl) {
             /* 70 */
-            nw = mlri(z, fnu, kode, n, cy, tol);
+            nw = mlri(z, fnu, kode, nn, cy, tol);
             if (nw < 0) {
                 nz = -1;
                 if (nw == -2) {
@@ -2788,7 +2788,11 @@ namespace amos {
         }
         /* 100 */
         if (nw > 0) {
-            return -1;
+            nz = -1;
+            if (nw == -2) {
+                nz = -2;
+            }
+            return nz;
         }
         nw = wrsk(z, fnu, kode, nn, cy, cw, tol, elim, alim);
         if (nw < 0) {
@@ -4296,7 +4300,7 @@ namespace amos {
         }
 
         if (p1 == 0.) {
-            p1 = std::complex<double>(0, 0);
+            p1 = std::complex<double>(tol, tol);
         }
 
         cy[n - 1] = p2 / p1;
@@ -4309,7 +4313,7 @@ namespace amos {
         cdfnu = fnu * rz;
 
         for (i = 2; i < (n + 1); i++) {
-            pt = cdfnu + t1 * rz * cy[k];
+            pt = cdfnu + t1 * rz + cy[k];
             if (pt == 0.0) {
                 pt = std::complex<double>(tol, tol);
             }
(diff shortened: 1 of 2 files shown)


dschmitz89:
Thanks for digging so deep into this difficult code.

Tests should go into [this folder](https://github.com/scipy/xsf/tree/main/tests/xsf_tests). You could mimic the approach from the other tests in that folder. With a little help from the LLM of your choice, it should hopefully not be too difficult (at least that was my experience as someone who never used ctest before). Do you have any specific questions?

include/xsf/amos/amos.h
```diff
@@ -2788,7 +2788,11 @@ namespace amos {
         }
         /* 100 */
         if (nw > 0) {
-            return -1;
+            nz = -1;
+            if (nw == -2) {
+                nz = -2;
+            }
+            return nz;
```
dschmitz89: The if condition in 2792 is unreachable as first we check for `nw>0` and then for `nw==-2`. Was this in the original Fortran source?

dschmitz89 (changes requested):

include/xsf/amos/amos.h
JaRoSchm: Thanks, of course you are right, I reverted this change.
Yes, this actually was this way in the Fortran source: https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zbinu.f#L99

JaRoSchm:
Thank you for the review! I added a test which fails on main and passes here. Mainly, I was worried about the size of the reference data. Here, I could reduce it to an array of length 15. However, in one of my other PRs, the length would be larger than 200 and I am not sure, if this is nice to put into the code.

When this is merged, I will update #92 and #93 and add tests to the same file.

include/xsf/amos/amos.h
```diff
@@ -2666,7 +2666,7 @@ namespace amos {
         dfnu = fnu + n - 1;
         if ((az <= 2.) || (az * az * 0.25 <= (dfnu + 1.0))) {
             /* GOTO 10 */
-            nw = seri(z, fnu, kode, n, cy, tol, elim, alim);
+            nw = seri(z, fnu, kode, nn, cy, tol, elim, alim);
```
JaRoSchm, line 2669: See https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zbinu.f#L26

include/xsf/amos/amos.h
```diff
@@ -2689,7 +2689,7 @@ namespace amos {
                 //
                 // MILLER ALGORITHM NORMALIZED BY THE SERIES
                 //
-                nw = mlri(z, fnu, kode, n, cy, tol);
+                nw = mlri(z, fnu, kode, nn, cy, tol);
```
JaRoSchm, line 2692: See https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zbinu.f#L66

include/xsf/amos/amos.h
```diff
@@ -2706,7 +2706,7 @@ namespace amos {
                 //
                 // ASYMPTOTIC EXPANSION FOR LARGE Z
                 //
-                nw = asyi(z, fnu, kode, n, cy, rl, tol, elim, alim);
+                nw = asyi(z, fnu, kode, nn, cy, rl, tol, elim, alim);
```
JaRoSchm, line 2709: See https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zbinu.f#L41

include/xsf/amos/amos.h
```diff
@@ -2760,7 +2760,7 @@ namespace amos {
         /* 60 */
         if (az <= rl) {
             /* 70 */
-            nw = mlri(z, fnu, kode, n, cy, tol);
+            nw = mlri(z, fnu, kode, nn, cy, tol);
```
JaRoSchm, line 2763: See https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zbinu.f#L66

include/xsf/amos/amos.h
```diff
@@ -4291,7 +4296,7 @@ namespace amos {
         }
 
         if (p1 == 0.) {
-            p1 = std::complex<double>(0, 0);
+            p1 = std::complex<double>(tol, tol);
```
JaRoSchm, line 4299: See https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zrati.f#L104

include/xsf/amos/amos.h
```diff
@@ -4304,7 +4309,7 @@ namespace amos {
         cdfnu = fnu * rz;
 
         for (i = 2; i < (n + 1); i++) {
-            pt = cdfnu + t1 * rz * cy[k];
+            pt = cdfnu + t1 * rz + cy[k];
```
JaRoSchm, line 4312: See https://github.com/scipy/scipy/blob/4edfcaa3ce8a387450b6efce968572def71be089/scipy/special/amos/zrati.f#L117

dschmitz89:
The translation errors look genuine. I am not so convinced about the outcome though. When I run your test data against scipy's main branch, the maximal relative error I get is in the order of 10 eps.

```python
import numpy as np
from mpmath import mp
from scipy.special import jv

mp.dps = 1000


n = 15
z = 70 - 0.7j

for n in range(1, n + 1):
    reference = complex(mp.besselj(n, mp.mpc(z)))
    scipy_result = jv(n, z)
    rel_error = abs(reference - scipy_result) / abs(reference)
    print(f"n={n}, mpmath.j{n}(z)={reference}, scipy.special.jv({n}, z)={scipy_result}, relative error={rel_error}")
``` 

Do your changes decrease the relative error even more?

JaRoSchm:
No, I expect no changes for scipy from this PR. Instead, this fixes a feature of the amos library which at the moment cannot be accessed through scipy (which is probably the reason why these bugs were not caught during the translation process). In the test file you can find the parameter `n` which is the number of consecutive orders of Bessel function returned. In the wrappers in `xsf/include/xsf/bessel.h` used by scipy this is always set to `n = 1`. However, it is planned to make this feature accessible for scipy, see the discussion in #50. So before approaching this, I want to fix the remaining translation bugs.

steppi (approved):
Looks good to me. Thanks @JaRoSchm and thanks @dschmitz89 for the review

PR #144, opened 2026-05-11
ENH: Updated and refactored Mathieu function implementation (part 1).

steppi:
#### Reference issue

Partially supersedes #99

This PR is based on @brorson's PR #99, which gives improved implementations of the Mathieu functions. Algorithmically, #99 is excellent work. This is a promised refactor in order to put the work in #99 into the form that would be most useful for xsf and SciPy. The primary departure is that this PR does not give full implementations of Mathieu functions; it only contains the core numerical kernels that would be useful in either a CPU implementation for SciPy or a GPU implementation for CuPy. The primary functions are for generating the recurrence matrices and for summing the Fourier series. Generating the Fourier coefficients requires finding eigenvectors of symmetric tridiagonal recurrence matrices. #99 used LAPACK's dstevd for this, which is out of scope for `xsf`. I do not rule out eventually writing a CPU+GPU capable symmetric tridiagonal eigenvector solver, but even then, it would still not be particularly useful for SciPy and CuPy to have full scalar implementations in `xsf`. For SciPy implementations, it is best to use the stateful-functor with in-ufunc caching pattern established here https://github.com/scipy/scipy/pull/24894, in order to allow reuse of the same computed fourier coefficients across varying values of `x`. For CuPy, we will want to use the gufunc expansion + reduction pattern proposed here https://github.com/cupy/cupy/issues/9839.  

The vision I'm coming to is that xsf will contain simple numerical kernels which will work on both CPU and GPU. If a function is simple enough for the same scalar implementation to work on both, then a full implementation belongs in xsf. Otherwise, xsf should contain the kernels that can be common to both platforms and CPU specific implementations can live in SciPy and GPU specific implementations in CuPy. 

This PR only contains kernels used for the characteristic value functions `mathieu_a` and `mathieu_b` and the Mathieu functions `mathieu_cem` and `mathieu_sem` from SciPy. I plan to address `mathieu_modcem1`, `mathieu_modcem2`, `mathieu_modsem1`, and `mathieu_modsem2` in an eventual follow-up. This is primarily to cut down the scope of work in this PR. There are also some redundant computations in the Fourier-Bessel summations from #99 that we can eliminate by pushing through #50, which is worth doing on its own as well.

I have used extensive templating to cut down on the duplicate code in #99. The implementations here should otherwise be completely faithful to #99, with the exception that I have added an `AngleUnits` policy template arg to the Fourier series summation so that the angular units can be taken to be either degrees or radians. For backwards compatibility, the companion PR to SciPy still takes arguments in degrees. There is an easy path forward to adding a kwarg to the SciPy functions to select between radians and degrees.

The tests here test only the recurrence matrix generation and the fourier summation. The tests for fourier summation use coefficients generated by the SciPy implementation and compare the evaluated sums to reference implementations of Mathieu functions from Mathematica. I've discovered the [Wolfram Engine Free Developers program](https://account.wolfram.com/access/wolfram-engine/free), and used the Python bindings to get reference values. This seems like something that will be useful for xsref too, which has been in a languishing state, but which I still have high hopes for.

I've removed the old Mathieu tests, which were self tests against the inaccurate SciPy code anyway. 

#### Additional information

After cleaning up the stuff adding BLAS and LAPACK to the dependencies, the rebuild of pixi.toml failed. This is because the cupy-tests environment was included in `solve-group = "default"`, but because the CuPy tests are Linux only and pull down things not available on other platforms, this caused solver errors. I simply removed the CuPy environment from that solve group.

#### AI Generation Disclosure

The test code in `xsf/tests/xsf_tests/test_mathieu.cpp` was largely AI generated with Google Gemini 3.1 Pro. I checked carefully that everything is correct. The implementation code is all good-old-fashioned artisanal hand-crafted code; this was too fun to delegate to AI.


steppi:
Oh wow, GPU CI actually ran after I updated the ubuntu runner to 24.04. Tests are failing, but I think will work if we update CuPy in the pixi.lock file. I'll make a separate PR for that.

steppi:
The formatting errors seem to be because the clang-format version was updated when I updated the pixi.lock file. Let's leave them here and fix them in another PR.

brorson:
Great to see this get merged in!

tests/xsf_tests/test_mathieu.cpp
```diff
@@ -0,0 +1,538 @@
+/* These tests are adapted from the test code written by Stuart Brorson for
+ * https://github.com/scipy/xsf/pull/99. It includes only tests for the
+ * kernels featured in xsf for generating the recurrence matrices and
+ * calculating Fourier sums. xsf does not contain full implementions of Mathieu
```
fbourgey:
```suggestion
 * calculating Fourier sums. xsf does not contain full implementations of Mathieu
```

tests/xsf_tests/test_mathieu.cpp
```diff
@@ -0,0 +1,538 @@
...
+    }
+
+    for (int i = 0; i < N; ++i) {
+        const double rel_error = xsf::extended_relative_error(D[i], D_expected[i]);
+        CAPTURE(i, D[i], D_expected[i], rtol, rel_error);
+        REQUIRE(rel_error <= rtol);
+    }
+}
+
+TEST_CASE("sum_fourier_series_even_even_q1_m0", "[mathieu][xsf_texts]") {
```
fbourgey:
```suggestion
TEST_CASE("sum_fourier_series_even_even_q1_m0", "[mathieu][xsf_tests]") {
```

tests/xsf_tests/test_mathieu.cpp
```diff
@@ -0,0 +1,538 @@
...
+        6.1152045985104173e-37, -5.2878909636082434e-40, 4.0787325607116287e-43, -2.8237182422819167e-46,
+        1.7643226442884998e-49, -9.9992520186900771e-53, 5.1636902188609567e-56, -2.4397830047988059e-59,
+        1.0587244524433085e-62,
+    };
+
+    std::mdspan AA_span(AA.data(), AA.size());
+
+    const std::vector<double> v = linspace(0.0, M_PI, 9);
+
+    /* Reference values computed with wolfram engine. */
```
fbourgey:
```suggestion
    /* Reference values computed with Wolfram Engine. */
```

tests/xsf_tests/test_mathieu.cpp
```diff
@@ -0,0 +1,538 @@
...
+                              0.0025865874071663215,   -4.3385838939211658e-05, 4.5372455146196267e-07,
+                              -3.2496292879186506e-09, 1.6958298519778925e-11,  -6.7394972996424881e-14,
+                              2.1085561518676298e-16,  -5.3296620523337814e-19, 1.1112098872910467e-21,
+                              -1.9473431483638467e-24};
+
+    std::mdspan AA_span(AA.data(), AA.size());
+
+    const std::vector<double> v = linspace(0.0, M_PI, 9);
+
+    /* Reference values computed with wolfram engine. */
```
fbourgey:
```suggestion
    /* Reference values computed with Wolfram Engine. */
```

I am stopping here but there are more below.

tests/xsf_tests/test_mathieu.cpp
```diff
@@ -0,0 +1,538 @@
...
+        xsf::mathieu::sum_fourier_series<Even, Even>(AA_span, v[i], out, out_diff);
+        double rel_error = xsf::extended_relative_error(out, expected[i]);
+        double rel_error_diff = xsf::extended_relative_error(out_diff, expected_diff[i]);
+        CAPTURE(m, q, out, out_diff, expected[i], expected_diff[i], rel_error, rel_error_diff);
+        REQUIRE(rel_error <= 1e-15);
+        REQUIRE(rel_error_diff <= 1e-15);
+    }
+}
+
+TEST_CASE("sum_fourier_series_even_even_q1_m2", "[mathieu][xsf_texts]") {
```
fbourgey:
```suggestion
TEST_CASE("sum_fourier_series_even_even_q1_m2", "[mathieu][xsf_tests]") {
```

I am stopping here but there are more below.

tests/xsf_tests/test_mathieu.cpp
```diff
@@ -0,0 +1,538 @@
...
+#include <xsf/third_party/kokkos/mdspan.hpp>
+
+auto constexpr Even = xsf::mathieu::Parity::Even;
+auto constexpr Odd = xsf::mathieu::Parity::Odd;
+
+TEST_CASE("make_matrix_ee", "[mathieu][xsf_tests]") {
+    int N = 6;
+
+    std::vector<double> D(N, 0.0);
+    std::vector<double> E(N, 0.0);
```
fbourgey: As this if the off-diagonal vector, should this be?

```suggestion
    std::vector<double> E(N-1, 0.0);
```

there are more below.

include/xsf/mathieu.h
```diff
@@ -1,228 +1,254 @@
...
+ * available at https://github.com/scipy/xsf/pull/99. It isolates out the parts
+ * which are fit for xsf as a library of simple numerical kernels which can be used
+ * on both CPU and GPU. Parts for which the approach used on CPU and GPU must
+ * necessarily be different will be left to SciPy and CuPy respectively. Hence
+ * the lack of the eigenvalue/eigenvector finding step here. The code has
+ * has been refactored to reduce code duplication, but is otherwise faithful
+ * to the original (with the exception that a policy template arg has been
+ * added to allow either degrees or radians to be used for angular arguments).
+ *
+ * Stuart Brorson's implementations follow prototypes written in MatLab and
```
fbourgey:
```suggestion
 * Stuart Brorson's implementations follow prototypes written in MATLAB and
```

include/xsf/mathieu.h
```diff
@@ -1,228 +1,254 @@
...
-        /* https://dlmf.nist.gov/28.2#E26 */
-        if (int_m % 2 == 0) {
-            return cem_cva(m, -q);
-        } else {
-            return sem_cva(m, -q);
+
+    /* Calculates the first element along the diagonal of the recurrence
+     * relation matrix as a function of the shape parameter ``q``.
+     *
+     * Templated on function parity (even vs odd mathieu functions) and
```
fbourgey:
```suggestion
     * Templated on function parity (even vs odd Mathieu functions) and
```

include/xsf/mathieu.h
```diff
@@ -1,228 +1,254 @@
...
-        int_m = (int)m;
-        if (q < 0) {
-            /* https://dlmf.nist.gov/28.2#E34 */
-            if (int_m % 2 == 0) {
-                sgn = ((int_m / 2) % 2 == 0) ? 1 : -1;
-                cem(m, -q, 90 - x, f, d);
-                csf = sgn * f;
-                csd = -sgn * d;
 
+    /* Sum the fourier series for computing angular Mathieu functions.
```
fbourgey:
```suggestion
    /* Sum the Fourier series for computing angular Mathieu functions.
```

I think we should use Fourier instead of fourier and Mathieu instead of mathieu. There are more below.

include/xsf/mathieu.h
```diff
@@ -1,228 +1,254 @@
...
+     *
+     * AngleUnits = AngleUnitPolicy::Radians or AngleUnitPolicy::Degrees.
+     *
+     * Inputs:
+     * X = A 1d mdspan view of the fourier coefficients.
+     * v = Angular argument, in either radians or degrees depending on the
+     * value of the AngleUnits template argument.
+     *
+     * Ouputs:
+     * out = Value of angular mathieu function for given fourier coefficients at angle v
```
fbourgey:
```suggestion
     * out = Value of angular mathieu function for given fourier coefficients at angle v.
```

include/xsf/mathieu.h
```diff
@@ -1,228 +1,254 @@
...
+    XSF_HOST_DEVICE inline int get_partial_sum_N(int m, double q) {
+        int N;
+
+        // This is sort of ad-hoc ...
+        if (q > 1.0) {
+            double qq = std::log10(q); // I need to use size of q to compute N.
+            N = m + 25 + 10 * qq;
         } else {
-            return cem_cva(m, -q);
+            N = m + 25;
```
fbourgey: If we think of the very negative $q$ case, should we not use `if (std::abs(q) > 1.0)` instead?

fbourgey (commented):
@steppi, this a first pass.

include/xsf/mathieu.h
```diff
@@ -1,228 +1,254 @@
...
+            } else {
+                xedp = xedp + td;
             }
+        } // for (k=(N-1) ...
+
+        // This makes sure the fcn has the right overall sign for q<0.
+        // Someday combine this with the above sums into the same for loop.
+        double x = 0.0;
+        for (decltype(N) l = 0; l < N; l++) {
+            x += X(l);
```
fbourgey: I understand that if `X` is an eigenvector, then so is `-X` but can you detail the rule "if the coefficients mostly sum negative, we flip the results" implemented here? I guess the coefficients of `X(l)` can have any sign in this framework (unless I missed something), so if the sum is very close to 0, can this be unstable? is there a better way?

steppi: This is a very good question and I think you've unearthed a potential bug. `dstevd` will return the eigenvectors normalized to have $l^2$ norm equal to 1, but the chosen sign is arbitrary, so we could either get the eigenvector $\mathbf{x}$ or $-\mathbf{x}$. These functions satisfy normalization conditions $\mathrm{cem}(m, q, 0) > 0$ and $\mathrm{sem}^{'}(m, q, 0) > 0$ (see details section below). $\mathrm{cem}(m, q, 0)$ is equal to the sum of the coefficients, so @brorson's idea here is to use this fact to get the correct sign for $\mathbf{x}$. But the correct normalization condition is $\mathrm{sem}^{'}(m, q, 0) > 0$ with $\mathrm{sem}(m, q, 0)$  identically 0. Rather than the sum of fourier coefficients, the value of this derivative at $x=0$ is

$$\mathrm{sem}^{'}(m, q, 0) = \frac{\mathrm{d}}{\mathrm{d}x}\left[\sum_{k}B_k \sin(kx)\right]_{x=0} = \sum_{k}kB_k$$

so I think to handle `FuncParity` odd, the above should be replaced with:

```C++
        if constexpr (FuncParity == Parity::Even) {
                x += X(l);
            } else {
                double r = sqrt_di<Parity::Odd, OrderParity>(l);
                x += r * X(l);
            }

```

@brorson, does that sound reasonable?

<details>
For $q = 0$ we have (https://dlmf.nist.gov/28.2#E29):

$$\mathrm{cem}(m, 0, x) = \cos(mx)$$
$$\mathrm{sem}(m, 0, x) = \sin(mx)$$

which implies that the sign conventions I mentioned above hold for $q = 0$. DLMF glosses over the convention for general $q$ by saying:

*the ambiguity of sign being resolved by ([28.2.29](https://dlmf.nist.gov/28.2#E29)) when q=0 and by continuity for the other values of q.*

It wasn't obvious to me at first, but suppose for instance there is some $q$ such that $\mathrm{cem}(m, q, 0) < 0$, by continuity, there must be some $q^{\star}$ for which $\mathrm{cem}(m, q^{\star}, 0) = 0$. Since we know $\mathrm{cem}(m, q^{\star}, x)$ is an even function, it follows that $\mathrm{cem}^{'}(m, q^{\star}, 0) = 0$. The only solution to Mathieu's equation satisfying these boundary conditions must be identically 0, from which we could derive a contradiction.

</details>

steppi: ~As for stability, the sum of coefficients is 0 at roots, and calculating these Mathieu functions seems to be inherently unstable near roots (or roots of the derivative). I found even the arbitrary precision Mathematica reference struggles near roots. We still have a major improvement over Zhang and Jin, so I think it's fine if there's still some rough edges.~

The above was nonsense. Sorry. The sum of coefficients is zero implies `mathieu_cem` is identically zero for all `x` since the function and it's derivative would be 0 at `x = 0` and $f(x) = 0$ is the only solution of Mathieu's equations satisfying these boundary conditions. 

I don't think this situation can occur for any values of `m` or `q`.

include/xsf/mathieu.h
steppi: I think you may be right. @brorson, can you comment on this?

steppi:
Thanks @fbourgey for the excellent review so far. I will fix all of the typos. I've tried to clarify the eigenvector sign situation, but also hope to hear from @brorson on this and on calculation of the number of fourier terms to take as a function of `m` and `q`.

include/xsf/mathieu.h
brorson: I think that's right.  The point here is that we need more terms in the sum when abs(q) > 1.   So, yes, if (std::abs(q) > 1.0) is correct.  Don't forget to use std::log10(std::abs(q)) on line 167 too.

include/xsf/mathieu.h
brorson: Yes, I think sum(k*B[k]) is reasonable.  Go ahead and change it.  I hope to have students run my test suite on the results at some point so if your idea doesn't work then we will pick it up.

brorson:
> but also hope to hear from @brorson on this and on calculation of the number of fourier terms to take as a function of `m` and `q`.

Done.  If I can do anything to help with the review process or merge let me know.  Once you have a stable branch please tell me and I can check it out and run some of my other tests on it.

steppi:
> > but also hope to hear from @brorson on this and on calculation of the number of fourier terms to take as a function of `m` and `q`.
> 
> Done. If I can do anything to help with the review process or merge let me know. Once you have a stable branch please tell me and I can check it out and run some of my other tests on it.

Thanks @brorson! I think this is getting into good shape now.  I'll ping you again if we need more help. I will also run out some more tests against Mathematica's implementation.

steppi:
@fbourgey, I think that should address everything from your first pass.

tests/xsf_tests/test_mathieu.cpp
```diff
@@ -0,0 +1,538 @@
...
+        CAPTURE(i, D[i], D_expected[i], rtol, rel_error);
+        REQUIRE(rel_error <= rtol);
+    }
+}
+
+TEST_CASE("make_matrix_eo", "[mathieu][xsf_tests]") {
+    int N = 6;
+
+    std::vector<double> D(N, 0.0);
+    std::vector<double> E(N, 0.0);
```
fbourgey:
```suggestion
    std::vector<double> E(N - 1, 0.0);
```

tests/xsf_tests/test_mathieu.cpp
```diff
@@ -0,0 +1,538 @@
...
+        CAPTURE(i, D[i], D_expected[i], rtol, rel_error);
+        REQUIRE(rel_error <= rtol);
+    }
+}
+
+TEST_CASE("make_matrix_oe", "[mathieu][xsf_tests]") {
+    int N = 6;
+
+    std::vector<double> D(N, 0.0);
+    std::vector<double> E(N, 0.0);
```
fbourgey:
```suggestion
    std::vector<double> E(N - 1, 0.0);
```

include/xsf/mathieu.h
```diff
@@ -1,228 +1,269 @@
+/* Kernels for computing Mathieu functions. This file does not contain complete
+ * implementations of the Mathieu functions, but only kernels which can be
+ * used within SciPy and CuPy to implement the Mathieu functions. One currently
+ * must supply an eigenvalue and eigenvector solver for symmetric tridiagonal
+ * matrices, such as dstevd from LAPACK, to construct a complete implementation.
+ *
+ * This code is based on code written by Stuart Brorson (github @brorson) which is
```
fbourgey:
```suggestion
 * This code is based on code written by Stuart Brorson (GitHub @brorson) which is
```

fbourgey (commented):
Minor things

brorson (approved):
Good by me.

Stuart

steppi:
@fbourgey, can you give this one last look?

tests/xsf_tests/test_mathieu.cpp
```diff
@@ -0,0 +1,538 @@
...
+        9.444966065193975e-17,
+        -0.509931239000559,
+        -0.6095378484410862,
+        -0.36076677721015726,
+        -1.1570532269097101e-16,
+    };
+
+    for (decltype(v.size()) i = 0; i < v.size(); i++) {
+        double out, out_diff;
+        xsf::mathieu::sum_fourier_series<Even, Even>(AA_span, v[i], out, out_diff);
```
fbourgey, line 177: Do we need a test for `AngleUnitPolicy::Degrees` as well?

include/xsf/mathieu_legacy.h
```diff
@@ -0,0 +1,230 @@
+/* Less accurate (but faster) legacy versions of Mathieu functions. */
```
fbourgey, line 1: what is the plan for this file? what about backward-compatibility? 

should we not keep old tests for smoke coverage?

fbourgey (approved):
This LGTM, modulo with last comments.

include/xsf/mathieu_legacy.h
steppi: This file remains for now because it's still being used for the modified (radial) mathieu functions, and as a large order fallback for `mathieu_cv`. Once the modified mathieu functions are refactored like this, and we figure out something else for the fallback, then we can get rid of the file. We can't really keep the old tests because they rely on computing the functions from end to end and `xsf` doesn't offer that capability. The full implementations for CPU live in SciPy and rely on kernels from here that the CuPy implementations will also use. All of those old test cases (using degrees as angular unit) remain in SciPy.

tests/xsf_tests/test_mathieu.cpp
steppi: Yeah, we can add some if you like.

steppi:
@fbourgey, I've added tests with degrees as angle units. If all looks good in CI I'm going to merge this.

PR #148, opened 2026-05-14
CI: hopefully get CuPy CI working and passing.

steppi:

I noticed here, https://github.com/scipy/xsf/pull/144#issuecomment-4454983210, that GPU CI actually ran in my latest push to #144. The change I made prior to this was updating the runner to Ubuntu 24.04. It seems that runner has CUDA version 13 on it, but the pixi.toml had has pinned to an earlier CuPy. This PR enforces at least CuPy 14 in the cupy-tests env. I had changed `cupy._core._create_ufunc` to use C++17 here (https://github.com/cupy/cupy/pull/9159/changes#diff-c9774fcecad94645bff3e07659c8eb24857a7bd1a08418651d430a42d3014f9c) which made it into the CuPy 14 release. C++17 is required for xsf, so CuPy 14 is the correct minimum CuPy version.
#### AI Generation Disclosure
No AI

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -3,6 +3,8 @@ environments:
   cupy-tests:
     channels:
     - url: https://prefix.dev/conda-forge/
+    options:
+      pypi-prerelease-mode: if-necessary-or-explicit
     packages:
       linux-64:
       - conda: https://prefix.dev/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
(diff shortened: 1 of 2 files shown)


dschmitz89:
Thanks @steppi ! Should we try to increase GPU test coverage?

steppi:
> Thanks @steppi ! Should we try to increase GPU test coverage?

Let's focus on having coverage for everything which is currently being used in CuPy, and incrementally adding coverage as things are added to CuPy.

PR #151, opened 2026-05-15
Fix xsf regressions

rgommers:
Multiple fixes for regressions I ran into in SciPy on Linux aarch64 with GCC 14.3.0 from conda-forge.

Individual commit messages contain more details. For the `nan`-related issues in `debugoptimized` build mode triggering with `_GLIBCXX_ASSERTIONS`, that is essentially the same as was fixed and tested for in gh-95. That testing looks incomplete, given that I found two more segfaults in the `scipy.special` test suite. Given the heavy use of `std::exp(complex`, there are likely more such issues.

_Draft PR because I'd like to test that these fixes are fine both in `xsf` and `scipy` CI first._

#### AI Generation Disclosure

Diagnosed and fixes drafted with heavy help from Claude Opus 4.7.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -20,7 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         environment: [tests-ci]
-        runs-on: [ubuntu-latest, macos-latest, windows-latest]
+        runs-on: [ubuntu-latest, ubuntu-24.04-arm, macos-latest, windows-latest]
 
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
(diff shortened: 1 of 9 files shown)


steppi:
Thanks for taking care of this @rgommers! This is great. Everything here looks sound.

steppi:
We have jobs that run CI in debug mode already and it seems we just need to add `ubuntu-24.04-arm` to the matrices here: https://github.com/scipy/xsf/blob/main/.github/workflows/tests.yml to get proper tests for this PR.

steppi:
Most of the linux_aarch64 failures just need tolerance bumps; it's just a case of not having tolerances tables for that specific platform. The `gamma` failures look real but unrelated. I don't think any of these failures are blockers for this PR. It' will be annoying that so many jobs will be red in CI, so I will try to add more platform specific tolerance files sooner than later.

rgommers:
> I don't think any of these failures are blockers for this PR.

This is blocking for the scipy 1.18.0rc1 next week, so it'd be good to get this in soon (ideally by tomorrow).

steppi:
Thanks Ralf! In it goes, everything looks good here.

PR #143, opened 2026-05-10
ENH: Improve j1 for large inputs

dschmitz89:
#### Reference issue
Towards https://github.com/scipy/scipy/issues/22706

#### What does this implement/fix?
Improve numerical stability for large inputs in $j1$ the same wa as #131 and #140.

#### Additional information
This PR again boils down to rewriting a trigonometrical expression. I used Wolfram Alpha to do that for me [here](https://www.wolframalpha.com/input?i=p+*+cos%28x-%283pi%2F4%29%29+%2B+w+*+q+*+sin%28x-%283pi%2F4%29%29) which yields the following screenshot:

image

Since this is the same pattern as #131 and #140 I went with a cutoff of 10 again. If requested, I can run a benchmark against mpmath to determine a suitable cutoff but I am quite confident that it will also yield something similar to 10.

#### AI Generation Disclosure
Copilot helped with test data generation and using them in the tests.

PR #152, opened 2026-05-16
ENH: use leading-order approximation for j1 at small |x|

dschmitz89:
#### Reference issue
Towards https://github.com/scipy/scipy/issues/22706

Fixes the $j1$ function in the far left tail using the suggestion $x/2$ for $x<sqrt(eps)$.

#### Additional information
I attach a crude testing script that establishes $sqrt(epsilon)$ as a suitable cutoff point.

<details><summary>Details</summary>
<p>

```python
import numpy as np
import matplotlib.pyplot as plt
from mpmath import mp
from scipy.special import j1

mp.dps = 1000

@np.vectorize
def mpmath_j1(x):
    x = mp.mpf(x)
    return float(mp.besselj(1, x))

min_subnormal = 2**(-1074)
x_values = np.logspace(np.log10(min_subnormal), 1, num=5000)
mpmath_results = mpmath_j1(x_values)
scipy_results = j1(x_values)
x_half = 0.5 * x_values
relative_errors_j1 = np.abs(mpmath_results - scipy_results) / np.abs(mpmath_results)
relative_errors_x_half = np.abs(mpmath_results - x_half) / np.abs(mpmath_results)
j1_better = relative_errors_j1 < relative_errors_x_half
print(f"First x value where j1 is better: {x_values[j1_better][0]}")
plt.figure(figsize=(10, 6))
#plt.loglog(x_values, relative_errors_j1, label='j1', color='blue')
plt.loglog(x_values, relative_errors_x_half, label='0.5 * x', color='green', linestyle='dashed')
#plt.title('Comparison of mpmath.j1 and scipy.special.j1')
plt.xlabel('x')
plt.ylabel('relative error')
plt.grid()
plt.legend()
plt.show()
``` 

</p>
</details> 

#### AI Generation Disclosure
AI was used to write the test cases. The actual test case values were manually created.

### DIFF
--- a/tests/xsf_tests/test_j1_left_tail.cpp
+++ b/tests/xsf_tests/test_j1_left_tail.cpp
@@ -1,0 +1,65 @@
+#include "../testing_utils.h"
+#include <xsf/cephes/j1.h>
+
+TEST_CASE("j1 left tail subnormal-input", "[j1][xsf_tests]") {
+    using test_case = std::tuple<double, double>;
+    // Reference values computed with mpmath with 1000 digits of precision:
+    //
+    // import math
+    // import mpmath as mp
+    // mp.mp.dps = 1000
+    // xs = np.logspace(np.log10(2**(-1074)), np.log10(np.finfo(np.float64).tiny), num=10)
+    // for x in xs:
+    //     print(x, mp.nstr(mp.besselj(1, x), 15))
+
+    auto [x, ref] = GENERATE(
+        test_case{5e-324, 0.0}, test_case{2.7e-322, 1.4e-322}, test_case{1.487e-320, 7.436e-321},
+        test_case{8.159e-319, 4.0795e-319}, test_case{4.4763326e-317, 2.2381663e-317},
+        test_case{2.4558778e-315, 1.2279389e-315}, test_case{1.3473833403e-313, 6.736916702e-314},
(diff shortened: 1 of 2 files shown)


fbourgey (commented):
I did a quick check and this looks good to me:

- The PR is a bit of a misnomer. I would avoid `far left tail` and use something like `ENH: use leading-order approximation for j1 at small |x|`.
- I guess the same patch could be used for j0, y1, etc?

dschmitz89:
> I did a quick check and this looks good to me:
> 
> * The PR is a bit of a misnomer. I would avoid `far left tail` and use something like `ENH: use leading-order approximation for j1 at small |x|`.

Good suggestion, done.

> * I guess the same patch could be used for j0, y1, etc?

Acc. to https://github.com/scipy/scipy/issues/22705 at least j0 does not suffer from this issue. y1 likely will, for higher orders I do not know. Still, I would not prioritise this highly as the precision issue mainly occurred in the subnormal regime. I am pretty sure that there are few real use cases for such tiny values. My main motivation for this PR was to close the SciPy issue as it is a quick (but small) win. A more important issue is the stability of y1 for large inputs (similar to #143 ) in my opinion.

tests/xsf_tests/test_j1_left_tail.cpp
steppi, line 1: The PR title / commit name were fixed, but it looks like "left-tail" language still snuck its way in here in the test name.

PR #153, opened 2026-05-18
MAINT: ensure zero sign is returned for tangd at multiples of 180 and for cotdg at multiples of 90

j-bowhay:
#### Reference issue

towards https://github.com/scipy/scipy/issues/20731

Ensures the following special values from the IEE spec are returned
image
I will have a follow-up PR for the poles.
#### AI Generation Disclosure

No AI

### DIFF
--- a/include/xsf/cephes/tandg.h
+++ b/include/xsf/cephes/tandg.h
@@ -103,7 +103,8 @@ namespace cephes {
             }
 
             /* modulo 180 */
-            x = x - 180.0 * std::floor(x / 180.0);
+            double k = std::floor(x / 180.0);
+            x = x - 180.0 * k;
             if (cotflg) {
                 if (x <= 90.0) {
                     x = 90.0 - x;
@@ -118,6 +119,9 @@ namespace cephes {
                 }
             }
             if (x == 0.0) {
+                if (static_cast<long long>(k) & 1) {
+                    sign *= -1;
+                }
                 return sign * 0.0;
             } else if (x == 45.0) {
                 return sign * 1.0;
--- a/tests/xsf_tests/test_trig.cpp
+++ b/tests/xsf_tests/test_trig.cpp
@@ -12,8 +12,64 @@ void check_zeros(double (*f)(double)) {
     }
 }
 
-TEST_CASE("sindg IEEE scipy/20731", "[sindg][xsf_tests]") { check_zeros(&xsf::sindg); }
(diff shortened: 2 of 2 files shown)


j-bowhay:
CI failures are unrelated

include/xsf/cephes/tandg.h
```diff
@@ -118,6 +119,9 @@ namespace cephes {
                 }
             }
             if (x == 0.0) {
+                if (static_cast<long long>(k) & 1) {
```
dschmitz89, line 122: I remember a lot of discussions about the portability of `long long`. Is it acceptable in xsf @steppi ?

include/xsf/cephes/tandg.h
```diff
@@ -103,7 +103,8 @@ namespace cephes {
             }
 
             /* modulo 180 */
-            x = x - 180.0 * std::floor(x / 180.0);
+            double k = std::floor(x / 180.0);
+            x = x - 180.0 * k;
```
dschmitz89, line 107: Is it possible to simplify here with `std::fmod`?

include/xsf/cephes/tandg.h
j-bowhay: There are other long long uses in xsf
https://github.com/scipy/xsf/blob/f7b85f505a94fd398024478d2e747b01dd2d7a6d/include/xsf/stats.h#L584

but I agree it would be good to double check with Albert

j-bowhay:
It is also worth noting the cotdg is not listed in the IEE standard but it is fairly straight forward to determine what special values it should return based on how tangd is defined. I also checked these against Julia

include/xsf/cephes/tandg.h
steppi: `long long` is an `int64` on all of the platforms we support and I think only isn't on niche exotic platforms. It's planned to  change to using fixed width ints everywhere in `xsf` but for now it's fine to use `long long` here.

include/xsf/cephes/tandg.h
j-bowhay: I don't think so since I need both the quotient and remainder, unless you had something else in mind?

steppi: Right, this looks fine as is.

dschmitz89:
Thanks @j-bowhay .

PR #156, opened 2026-05-19
config: emit __host__ __device__ on HIP-clang too

gpinkert:
Under HIP-clang's single-source compilation model, every helper that might be reached from a __global__ kernel must carry an explicit __device__ qualifier; without it the compiler aborts with

    error: call to __host__ function from __global__ function

while listing the (post-preprocessor) signature

    inline double digamma(double z) { ... }

as the only candidate. The xsf headers declare the helper as

    XSF_HOST_DEVICE inline double digamma(double z) { ... }

but XSF_HOST_DEVICE expands to nothing on every non-__CUDACC__ target, so under HIP it disappears and the function ends up host-only.

NVCC has long silently relaxed this rule for unannotated inline functions, so the upstream macro definition was never a problem on CUDA. HIP-clang -- like recent (post-9.0) NVCC '--strict-host-device' mode -- enforces the boundary, and any consumer that pulls xsf::digamma, xsf::wright_bessel, xsf::cephes::ellie, etc. into a kernel on the HIP backend trips over it.

Add a HIP-clang arm to the existing #ifdef __CUDACC__ ladder that keeps the bare-STL include set used by the generic host path (pulling in <cuda/std/...> here would conflict with libstdc++) but redefines XSF_HOST_DEVICE to '__host__ __device__' so every XSF_HOST_DEVICE- prefixed helper picks up the device qualifier HIP-clang requires.

Reproducer (CuPy, ROCm 7.2 on gfx90a):

    import cupy
    from cupyx.scipy.special import digamma
    digamma(cupy.asarray([1.5, 2.5]))

Before the patch, the kernel compile aborts with the "call to __host__ function from __global__ function" error above on every helper xsf marks XSF_HOST_DEVICE. With the patch the kernel compiles and runs unchanged.

This commit is for CuPy enablment on AMD hardware.

### DIFF
--- a/include/xsf/config.h
+++ b/include/xsf/config.h
@@ -235,7 +235,12 @@ using cuda::std::uint64_t;
 } // namespace std
 
 #else
+#if defined(__HIPCC__) || defined(__HIP__)
+// HIP-clang strictly enforces host/device call boundaries.
+#define XSF_HOST_DEVICE __host__ __device__
+#else
 #define XSF_HOST_DEVICE
+#endif
 
 #include <algorithm>
 #include <cassert>


steppi:
Thanks @gpinkert! I don't have AMD hardware to test on yet but will take your word that this is sufficient to get things working. The CI failures for windows and linux_aarch64 are unrelated. 

Irwin was the one who started the pattern of injecting parts of `cuda::std` into `std` by writing wrappers. I can't remember the details behind why we went with that pattern, but it works in CuPy when using CUDA, and we've stuck with it. I'm afraid I don't currently know the details of how `xsf/config.h` interacts with CuPy as well as I should. When compiling with `HIP` in CuPy, how is it that the standard generic host path with the bare STL include set works on device? Do I understand correctly that it's a consequence of HIP-clang's single source compilation model that the following includes

```C++
#include <algorithm>
#include <cassert>
#include <cmath>
#include <complex>
#include <cstddef>
#include <cstdint>
#include <iterator>
#include <limits>
#include <math.h>
#include <tuple>
#include <type_traits>
#include <utility>
```

will get the relevant host stdlib headers when compiling for host, and the relevant device headers when compiling for device?  Making it easier to write code that works on host and device like we are aiming for with `xsf`. If so, that sounds really cool.

gpinkert:
Hey, thanks for the fast response. Making sure that I understand the question - you're asking if we can just write something like this and it just works with finding the host and device overloads for relevant functions (excuse the hastily written pseudocode):

```c++
#include <algorithm>

template<typename T> 
__global__ void demo_kernel(T* in)
{
   
    if (!std::isnan(in[0]) 
    {
        // Do some op
    }
    // else explode
}

int main()
{
    float[] val = 0.0f;
    if (std::is_finite(val))
    {
         launchHipKernel(demo_kernel, val);
    }
}
``` 

The answer is yes - see this working example: https://godbolt.org/z/jzzndx16G. 

Technically though: it's not actually two different headers. It's the same libstdc++ headers in both compilation passes, but clang ships a small wrapper at  `<clang-resource>/include/cuda_wrappers/<header>` that uses `#pragma clang force_cuda_host_device` around an `#include_next`, which retroactively decorates the libstdc++ functions as `__host__ __device__`. So the same std::isnan you'd get on a regular x86_64 build becomes device-callable inside `__global__` kernels via that wrapper, without any user-side annotation.

include/xsf/config.h
steppi, line 262: I included `XSF_ASSERT` here to replace existing asserts that were behind a debug flag  in the `scipy.special` codebase, under the assumption that this compilation branch is CPU only. The corresponding section of the `#ifdef __CUDACC__` branch unconditionally makes `XSF_ASSERT` a no-op. 

To be strictly correct. It seems like we'd also want `XSF_ASSERT` to be a no-op when compiling under `HIP`, so perhaps it makes sense to put a no-op definition of `XSF_ASSERT` in the `#if defined(__HIPCC__) || defined(__HIP__)` conditional branch, and wrap the highlighted section here in `#ifndef XSF_ASSERT`.

steppi (approved):
Thanks @gpinkert! I have one minor suggestion, but this looks good.

steppi:
> Technically though: it's not actually two different headers. It's the same libstdc++ headers in both compilation passes, but clang ships a small wrapper at `<clang-resource>/include/cuda_wrappers/<header>` that uses `#pragma clang force_cuda_host_device` around an `#include_next`, which retroactively decorates the libstdc++ functions as `__host__ __device__`. So the same std::isnan you'd get on a regular x86_64 build becomes device-callable inside `__global__` kernels via that wrapper, without any user-side annotation.

This sounds really cool. I wasn't aware of that. The two headers idea I had was to have say an `<algorithm>` header which itself conditionally includes the C++ header from the stdlib or a device header. I'm curious how it smooths over incompatibilities between what's available on host and device.

steppi:
@gpinkert. I'm going to go ahead and merge this since the thing with `XSF_ASSERT` doesn't seem that important. I see that you work at AMD. Do you have any recommendations for the best path to add a ROCm CI job here?

PR #157, opened 2026-05-19
digamma: handle non-finite complex inputs before std::abs

gpinkert:
xsf::digamma(std::complex<double>) starts with

    double absz = std::abs(z);

and then dispatches every subsequent branch on absz comparisons. Under libstdc++ the generic __complex_abs template (selected whenever __builtin_cabs is not picked -- in particular whenever _GLIBCXX_USE_C99_COMPLEX is 0, which is the default that clang's cuda_wrappers/complex pins for both CUDA and HIP) applies the standard scale-by-max-magnitude algorithm

    template<typename _Tp>
    inline _Tp __complex_abs(const complex<_Tp>& z) {
        _Tp x = z.real();
        _Tp y = z.imag();
        const _Tp s = std::max(abs(x), abs(y));
        if (s == _Tp()) return s;
        x /= s;                            // inf/inf == NaN
        y /= s;                            // 0  /inf == 0
        return s * sqrt(x*x + y*y);        // inf * sqrt(NaN) == NaN
    }

That algorithm is correct for every finite input, but produces NaN when an operand is already infinite (inf/inf == NaN).

Cascade for digamma(+inf + 0j):

    absz = std::abs(z) = NaN
    if (absz < 0.5)             ->  NaN < 0.5  = false (skip)
    if (absz > smallabsz)       ->  NaN > 16   = false (skip!)
    else if (z.real() >= 0.0)   ->  enter
        n = trunc(16 - NaN) + 1                = NaN
        z + n                                  = NaN + 0j
        digamma_asymptotic_series(NaN + 0j)    = NaN + NaN*i
        return NaN + NaN*i

So digamma(+inf + 0j) returns NaN+NaNj instead of the expected +inf + 0j. The same routing failure affects -inf + finite*j (which silently returns 0+0j because every branch's pole / reflection / recurrence guard also false-routes on NaN) and nan + finite*j.

Fix: short-circuit non-finite inputs before reaching std::abs(z) and return the analytic limit directly:

  * any NaN component   -> NaN + NaN*j
  * imaginary infinity  -> NaN + NaN*j  (no finite limit; dense poles)
  * +inf + finite*j     -> +inf + 0j    (psi -> +inf on positive real)
  * -inf + finite*j     -> NaN + NaN*j  (limit of poles, matches
                                         the existing pole branch)

The patch is a strict improvement even for builds that go through __builtin_cabs (and therefore return inf correctly for absz): on those builds, +inf + 0j happens to land in the asymptotic branch whose early-return for non-finite z masks the issue, but nan + 0j still silently returns 0+0j today; this patch makes it return NaN+ NaN*j instead.

Zero overhead on the hot path: two isfinite tests on entry, then the original code.

Reproducer (CuPy, ROCm 7.2 on gfx90a):

    import cupy
    from cupyx.scipy.special import digamma
    import numpy as np
    digamma(cupy.asarray([np.inf + 0j], dtype=cupy.complex128))

Before:  array([nan+nanj])
After:   array([inf+0.j])

Adds tests/xsf_tests/test_digamma.cpp with two TEST_CASEs:

  * "digamma non-finite complex inputs" -- one SECTION per analytic- limit class above, including the explicit (+inf, 0) -> (+inf, 0) case that motivated the patch.

  * "digamma non-finite enumeration" -- enumerates every (re, im) combination of {finite, +inf, -inf, NaN}, asserts call-determinism (two calls return bit-identical results) and asserts the result matches the analytic-limit table. Without the patch this test fails on every infinite input.

### DIFF
--- a/include/xsf/digamma.h
+++ b/include/xsf/digamma.h
@@ -34,6 +34,16 @@ namespace detail {
     // Value of the negative root
     constexpr double digamma_negrootval = 7.2897639029768949e-17;
 
+    // Limit of digamma(z) for non-finite complex z.
+    // Precondition: !std::isfinite(z.real()) || !std::isfinite(z.imag()).
+    template <typename T>
+    XSF_HOST_DEVICE std::complex<T> digamma_nonfinite_limit(std::complex<T> z) {
+        constexpr T qnan = std::numeric_limits<T>::quiet_NaN();
+        if (std::isnan(z.real()) || !std::isfinite(z.imag()) || z.real() <= T{0})
+            return {qnan, qnan};
+        return {std::numeric_limits<T>::infinity(), T{0}};
+    }
+
     template <typename T>
     XSF_HOST_DEVICE T digamma_zeta_series(T z, double root, double rootval) {
         T res = rootval;
@@ -144,6 +154,12 @@ XSF_HOST_DEVICE inline std::complex<double> digamma(std::complex<double> z) {
      * - If |z| is small, use a recurrence relation to make |z| large
      * enough to use the asymptotic series.
      */
+
+    // Non-finite z: std::abs(complex) can return NaN, false-routing dispatch.
+    if (!std::isfinite(z.real()) || !std::isfinite(z.imag())) {
+        return detail::digamma_nonfinite_limit(z);
+    }
+
     double absz = std::abs(z);
     std::complex<double> res = 0;
     /* Use the asymptotic series for z away from the negative real axis
--- a/tests/xsf_tests/test_digamma.cpp
+++ b/tests/xsf_tests/test_digamma.cpp
@@ -1,0 +1,105 @@
+#include "../testing_utils.h"
+#include <xsf/digamma.h>
+
+#include <array>
+#include <catch2/catch_template_test_macros.hpp>
+#include <catch2/matchers/catch_matchers.hpp>
+#include <catch2/matchers/catch_matchers_floating_point.hpp>
+#include <complex>
+#include <limits>
+
+using Catch::Matchers::IsNaN;
+using Catch::Matchers::WithinAbs;
+
+// Regression: digamma must short-circuit non-finite complex inputs
+// before std::abs(z), which can return NaN for inf operands under
+// libstdc++ and false-route every subsequent dispatch comparison.
+
+TEMPLATE_TEST_CASE("digamma non-finite complex inputs", "[digamma][xsf_tests][nonfinite]", float, double) {
+    using T = TestType;
+    using cd = std::complex<T>;
+    constexpr T inf = std::numeric_limits<T>::infinity();
+    constexpr T qnan = std::numeric_limits<T>::quiet_NaN();
+
+    SECTION("+inf real -> +inf + 0j") {
+        for (T im : std::array<T, 4>{T{0}, T{1}, T{-1}, T{1e30}}) {
+            cd r = xsf::digamma(cd{inf, im});
+            CAPTURE(im, r);
+            REQUIRE(r.real() == inf);
+            REQUIRE(r.imag() == T{0});
+        }
+    }
+
+    SECTION("-inf real -> NaN") {
+        for (T im : std::array<T, 3>{T{0}, T{1}, T{-1e30}}) {
+            cd r = xsf::digamma(cd{-inf, im});
+            CAPTURE(im, r);
+            REQUIRE_THAT(r.real(), IsNaN());
+            REQUIRE_THAT(r.imag(), IsNaN());
+        }
+    }
+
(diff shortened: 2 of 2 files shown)


steppi:
Thanks @gpinkert. These kinds of issues with compilers that don't support C99 annex G complex numbers has been a recurring problem in `xsf`. It's the reason behind the windows failures you may have noticed in the CI of your other PR #156.  This kind of short-circuiting is exactly what I had in mind for fixing it, but I haven't had bandwidth to get to it.

The CI failures here look genuine and the new test file you added seems like it needs some fixes, but I agree with the overall approach taken here.

gpinkert:
gotcha. I will take a closer look sometime later this week

gpinkert:
If you have a plan for the short circuiting, maybe I could help pick up any slack?

gpinkert:
@steppi to clarify, you would be ok to merge if I just fix the test errors?

steppi:
> @steppi to clarify, you would be ok to merge if I just fix the test errors?

Essentially yes. I will give this another look before merging, but at a glance at least, things look correct. The test errors are the blocker right now.

gpinkert:
@steppi it seems like there are less test failures than my previous PR which was merged. Let me know if there is anything else I need to do.

steppi:
Thanks @gpinkert. The remaining failures are the known ones that occur because we don't have arm specific tolerance files yet. I'm going to add github workflows to create and update the tolerance files for all of the platforms we test, but until then we just ignore the known failures.

include/xsf/digamma.h
```diff
@@ -34,6 +34,20 @@ namespace detail {
     // Value of the negative root
     constexpr double digamma_negrootval = 7.2897639029768949e-17;
 
+    // Limit of digamma(z) for non-finite complex z.
+    // Precondition: !std::isfinite(z.real()) || !std::isfinite(z.imag()).
+    template <typename T>
+    XSF_HOST_DEVICE inline std::complex<T> digamma_nonfinite_limit(std::complex<T> z) {
```
steppi:
```suggestion
    XSF_HOST_DEVICE std::complex<T> digamma_nonfinite_limit(std::complex<T> z) {
```
Just a style thing, but since `inline` is redundant for templates (they already avoid the one definition rule) we try to leave the `inline` keyword off of templates.

include/xsf/digamma.h
```diff
@@ -34,6 +34,20 @@ namespace detail {
...
+        const T qnan = std::numeric_limits<T>::quiet_NaN();
```
steppi:
```suggestion
        constexpr T qnan = std::numeric_limits<T>::quiet_NaN();
```

It shouldn't actually make a difference in generated assembly in this case, but for consistency we like to make aliases like this `constexpr`.

include/xsf/digamma.h
```diff
@@ -34,6 +34,20 @@ namespace detail {
...
+        if (std::isnan(z.real()) || std::isnan(z.imag()))
+            return {qnan, qnan};
+        if (std::isinf(z.imag()))
+            return {qnan, qnan};
+        if (z.real() > T{0})
+            return {std::numeric_limits<T>::infinity(), T{0}};
+        return {qnan, qnan};
```
steppi:
```suggestion
        if (std::isnan(z.real()) || !std::isfinite(z.imag()) || z.real() <= T{0}) 
            return {qnan, qnan}; 
        return {std::numeric_limits<T>::infinity(), T{0}};
```

I find this more easy to follow when done with fewer branches

steppi (changes requested):
Thanks @gpinkert! This looks good from a mathematical perspective; I have a few minor suggestions related to code style. Sorry for not having a style-guide posted yet

include/xsf/digamma.h
gpinkert: No worries, I will whip my claude agent into shape

include/xsf/digamma.h
gpinkert: I think it might, constexpr can reduce the used registers right? If a variable get's it's address taken somewhere, or the compiler can't reason that the variable is used outside of a constexpr expression and gives it storage.

include/xsf/digamma.h
gpinkert: I agree, will fix

gpinkert:
@steppi should be good to re-review

include/xsf/digamma.h
steppi: Good point, that sounds right in principle. I guess that's why it's good practice to always use `constexpr` for such cases, so we don't even have to think about details like that. In this particular case I'd expect any reasonable compiler set at a decent optimization level to get this right though.

steppi (approved):
Looks great. Thanks @gpinkert!

PR #133, opened 2026-04-25
ENH: add nrdtrimn and nrdtrisd to stats.h

CodeHermit17:
Part of #127. References #128.

Adds `nrdtrisd` to `ndtri_reparametrizations.h` alongside `nrdtrimn` from #128. The function signature and design follows the discussion in those issues.

Includes `double` and `float` overloads, basic tests, and NaN input validation.

PR #160, opened 2026-05-22
ENH: add gdtria and gdtrix

CodeHermit17:
#### Reference issue
Part and continuation of #127.

Adds `gdtrix` and `gdtria` to `cdflib.h` as `XSF_HOST_DEVICE` inline functions.
Both functions include input validation and domain error handling. Note that `gdtria` now correctly returns NaN for `x <= 0`, and `gdtrix` now correctly returns NaN for `a < 0`, improving on the existing SciPy implementation, which does not validate negative x values.

Inclusion of input validation and domain error handling is based on https://people.sc.fsu.edu/~jburkardt/f_src/cdflib/cdflib.f90 .
I've added them in `cdflib.h` as `gdtrib` was implemented there.
Followed the testing convention mentioned in PR #133.
#### Additional information
No float overloads added, consistent with `gdtrib` in the same file.

#### AI Generation Disclosure
No AI tools used.


dschmitz89 (changes requested):
Thanks again @CodeHermit17 .

Two comments: we should stay backwards compatible, so all inputs must stay allowed as they are now in SciPy. I agree that negative parameters do not make much sense but we still need an official depreciation period. So please remove those changes.

Just for consistency, let's add the float overload here also.

tests/xsf_tests/test_gdtria_gdtrix.cpp
```diff
@@ -0,0 +1,354 @@
...
+
+        // for p, b, x in zip(p_s, b_s, x_s):
+        //     output = special.gdtria(p, b, x)
+        //     print(f"test_case{{{p}, {b}, {x}, {output}}}")
+        using test_case = std::tuple<float, float, float, float>;
+        auto [p, b, x, expected] = GENERATE(
+            test_case{0.2327893078327179, 8.561944961547852, -4.895352840423584, -1.2905455827713013},
+            test_case{0.3204231858253479, 6.0560503005981445, -4.048694610595703, -1.157485842704773},
+            test_case{0.7914181351661682, 9.326684951782227, -4.4186577796936035, -2.6354193687438965},
+            test_case{0.6727295517921448, 7.275335311889648, -4.788415908813477, -1.7096890211105347},
```
dschmitz89: These are the same test data as in the test before. Can we not combine them into one test that tests double and float precision to avoid repetition? The same holds for the tests for gdtrix.

tests/xsf_tests/test_gdtria_gdtrix.cpp
```diff
@@ -0,0 +1,354 @@
...
+            test_case{1.0, 1.0, 0.2, std::numeric_limits<float>::infinity()},   // p == 1
+            test_case{-0.1, 1.0, 0.2, std::numeric_limits<float>::quiet_NaN()}, // p < 0
+            test_case{1.1, 1.0, 0.2, std::numeric_limits<float>::quiet_NaN()},  // p > 1
+            test_case{0.5, 1.0, 0.0, std::numeric_limits<float>::quiet_NaN()}   // x == 0
+        );
+        const float rtol = 100 * std::numeric_limits<float>::epsilon();
+        const auto output = xsf::gdtria(p, b, x);
+        const auto rel_error = xsf::extended_relative_error(output, expected);
+        CAPTURE(p, b, x, output, expected, rtol, rel_error);
+        REQUIRE(rel_error <= rtol);
```
dschmitz89: All reference values here are either nan or inf, so the relative error is not meaningful. Lets simply do a strict comparision (`std::isnan`/`std::isinf` plus sign check).

steppi: Actually, for `extended_relative_error` this is meaningful and should be left unchanged. This is what the "extended" in `extended_relative_error` means. See https://github.com/scipy/xsf/blob/main/include/xsf/fp_error_metrics.h.

dschmitz89 (approved):

include/xsf/cdflib.h
```diff
@@ -1,13 +1,31 @@
-
 #pragma once
```
dschmitz89, line 1:
```suggestion

#pragma once
```

Lets leave this as it was.

include/xsf/cdflib.h
```diff
@@ -4,10 +4,29 @@
...
+    if (x == 0) {
+        return std::numeric_limits<double>::quiet_NaN();
+    }
+    if ((b == 0) & (p == 0)) {
+        if (std::isinf(x) && (x > 0)) {
+            return std::numeric_limits<double>::quiet_NaN();
+        }
+        return 0.0;
+    }
+    return gammaincinv(b, p) / x;
```
fbourgey, line 22: https://github.com/scipy/xsf/blob/676ceb53bd0e17b57b872919b964deb1112724b7/include/xsf/gamma.h#L19

is not marked XSF_HOST_DEVICE. Will this break for CuPy/GPU consumers? If so, I think we should add GPU compatibility to `gammaincinv`. I guess we should do the same for the remaining gamma* functions.

cc @steppi to confirm

include/xsf/cdflib.h
```diff
@@ -4,10 +4,29 @@
...
 #include "tools.h"
 
 namespace xsf {
 
+XSF_HOST_DEVICE inline double gdtria(double p, double b, double x) {
+
+    if (x == 0) {
+        return std::numeric_limits<double>::quiet_NaN();
+    }
+    if ((b == 0) & (p == 0)) {
```
fbourgey:
```suggestion
    if ((b == 0) && (p == 0)) {
```

I'd use logical `&&` for consistency, though `&` works for booleans.

include/xsf/cdflib.h
steppi: Indeed. I added the annotation here, https://github.com/scipy/xsf/pull/164/changes#diff-ad5b10f5a55a8b34233dfece4768d5c47b29a8a439e26bdce782c2b47c5a65ef, so maybe let's just get that PR in first.

include/xsf/cdflib.h
steppi: And not just consistency, the bitwise `&` won't short-circuit and runs an extra comparison if the first is false.

include/xsf/cdflib.h
dschmitz89: Ouch, thanks for spotting this. I had assumed that most of cephes was GPU ready.

steppi: Indeed most of cephes is, but the annotations got left out of that top levell `xsf/gamma.h` header for some reason, and we didn't notice because CuPy doesn't use it yet.

fbourgey:
LGTM, thanks.

PR #162, opened 2026-05-24
BUG: avoid spurious overflow warning in j0/y0 for tiny x

### LINKED ISSUE #25199: BUG: special: y0(1e-200) generates an incorrect overflow warning when enabled.
### Describe your issue.

When overflow warnings are enabled, `y0(1e-200)` generates a warning, even though the result is correct to machine precision.

In the ipython session below, you can see that the return value, `-293.2480438468798`, has not overflowed, and in fact it agrees exactly with the result computed with `mpmath`.

### Reproducing Code Example

```python
In [1]: import scipy

In [2]: scipy.__version__
Out[2]: '1.19.0.dev0+git20260523.44a7d64'

In [3]: from scipy.special import y0, seterr

In [4]: old_state = seterr(overflow='warn')

In [5]: x = 1e-200

In [6]: y0(x)
<ipython-input-6-24b7e7629dab>:1: SpecialFunctionWarning: scipy.special/y0: (overflow) floating point overflow
  y0(x)
Out[6]: np.float64(-293.2480438468798)

In [7]: from mpmath import mp

In [8]: mp.dps = 75

In [9]: float(mp.bessely(0, x))
Out[9]: -293.2480438468798
```

<details>
<summary>

### SciPy/NumPy/Python version and system information

</summary>

```shell
Build Dependencies:
  blas:
    cython blas ilp64: false
    detection method: pkgconfig
    found: true
    has ilp64: false
    include directory: /usr/include/x86_64-linux-gnu/openblas-pthread/
    lib directory: /usr/lib/x86_64-linux-gnu/openblas-pthread/
    name: openblas
    openblas configuration: USE_64BITINT=0 DYNAMIC_ARCH=1 DYNAMIC_OLDER=1 NO_CBLAS=
      NO_LAPACK= NO_LAPACKE=1 NO_AFFINITY=1 USE_OPENMP=0 generic MAX_THREADS=64
    pc file directory: /usr/lib/x86_64-linux-gnu/pkgconfig
    version: 0.3.26
  lapack:
    detection method: pkgconfig
    found: true
    has ilp64: false
    include directory: /usr/include/x86_64-linux-gnu/openblas-pthread/
    lib directory: /usr/lib/x86_64-linux-gnu/openblas-pthread/
    name: openblas
    openblas configuration: USE_64BITINT=0 DYNAMIC_ARCH=1 DYNAMIC_OLDER=1 NO_CBLAS=
      NO_LAPACK= NO_LAPACKE=1 NO_AFFINITY=1 USE_OPENMP=0 generic MAX_THREADS=64
    pc file directory: /usr/lib/x86_64-linux-gnu/pkgconfig
    version: 0.3.26
  pybind11:
    detection method: config-tool
    include directory: unknown
    name: pybind11
    version: 3.0.4
Compilers:
  c:
    commands: cc
    linker: ld.bfd
    name: gcc
    version: 13.3.0
  c++:
    commands: c++
    linker: ld.bfd
    name: gcc
    version: 13.3.0
  cython:
    commands: cython
    linker: cython
    name: cython
    version: 3.2.4
  fortran:
    commands: gfortran
    linker: ld.bfd
    name: gcc
    version: 13.3.0
  pythran:
    include directory: ../../../../../py3.14.5/lib/python3.14/site-packages/pythran
    version: 0.18.1
Machine Information:
  build:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
  cross-compiled: false
  host:
    cpu: x86_64
    endian: little
    family: x86_64
    system: linux
Python Information:
  path: /home/warren/py3.14.5/bin/python3.14
  version: '3.14'
```

</details>

trustdemons05:
### Title
BUG: cephes: avoid spurious FP exception in j0/y0 for tiny x
### Description
For very small x (< 3e-8), `x^2/4` is below machine epsilon so
`1 - x^2/4` rounds to 1.0. Avoid computing `x*x`, which also
prevents spurious underflow exceptions for extreme values near
`sqrt(DBL_MIN)`. Use limiting forms directly:
- `j0(x) ≈ 1.0`  error is O(x²), far below machine precision
- `y0(x) ≈ YP[7]/YQ[6] + (2/π)·log(x)` — the rational approximation
  reduces to its constant term.
This preserves the exact numerical result while avoiding the underflowing
computation. The threshold 3e-8 is chosen because `x^2/4 < ε_mach` for
all x < 3e-8.
Includes GENERATE-based tests that verify:
- Correct values for j0 and y0 at tiny x
- No FE_UNDERFLOW exception is raised
- y0 at 1e-100 and 2e-154 is finite, negative, and exception-free
cc @steppi
Fixes scipy/scipy#25199

I used an AI coding assistant (opencode) to help draft parts of this fix, including the threshold constant extraction and test scaffolding , Also used for text formatting. All code was reviewed, understood, and verified by me before submission.

### DIFF
--- a/include/xsf/cephes/j0.h
+++ b/include/xsf/cephes/j0.h
@@ -168,9 +168,7 @@ namespace cephes {
 
         if (x <= 5.0) {
             if (x < 3e-8) {
-                /* For x < 3e-8, 1 - x^2/4 rounds to 1.0 in double precision
-                 * (x^2/4 is below machine epsilon). Avoid computing x*x entirely.
-                 */
+                /* x^2/4 is below machine epsilon — 1 - x^2/4 rounds to 1.0. */
                 return (1.0);
             }
             if (x < 1.0e-5) {
@@ -217,10 +215,8 @@ namespace cephes {
                 set_error("y0", SF_ERROR_DOMAIN, NULL);
                 return std::numeric_limits<double>::quiet_NaN();
             }
-            /* When x < 3e-8, x*x/4 is below machine epsilon for double precision.
-             * Avoid computing x*x entirely by using the limiting form directly.
-             */
             if (x < 3e-8) {
+                /* x*x/4 below machine epsilon — use limiting form directly. */
                 w = detail::j0_YP[7] / detail::j0_YQ[6];
                 w += M_2_PI * std::log(x);
                 return (w);
--- a/include/xsf/cephes/j0.h
+++ b/include/xsf/cephes/j0.h
@@ -105,7 +105,7 @@ namespace cephes {
 
     namespace detail {
 
-        inline constexpr double double_min = 1.4916681462400413e-154;
+        inline constexpr double sqrt_double_min = 1.4916681462400413e-154;
 
         constexpr double j0_PP[7] = {
             7.96936729297347051624E-4, 8.28352392107440799803E-2, 1.23953371646414299388E0,  5.44725003058768775090E0,
@@ -170,11 +170,11 @@ namespace cephes {
 
         if (x <= 5.0) {
             if (x < 1.0e-5) {
-                /* For x < double_min (~1.49e-154), x*x underflows to zero and raises a
+                /* For x < sqrt_double_min (~1.49e-154), x*x underflows to zero and raises a
                  * spurious floating-point exception. j0(x) = 1 - x^2/4 + ...
                  * is 1.0 to machine precision when x*x underflows.
                  */
-                if (x < detail::double_min) {
+                if (x < detail::sqrt_double_min) {
                     return (1.0);
                 }
                 z = x * x;
@@ -220,12 +220,12 @@ namespace cephes {
                 set_error("y0", SF_ERROR_DOMAIN, NULL);
                 return std::numeric_limits<double>::quiet_NaN();
             }
-            /* When x < double_min (~1.49e-154), x*x underflows to zero, which raises a
+            /* When x < sqrt_double_min (~1.49e-154), x*x underflows to zero, which raises a
              * spurious floating-point exception. Avoid the underflow by using
              * the limiting form directly: for tiny x, j0(x) ≈ 1 and the
              * rational approximation reduces to its constant term.
              */
-            if (x < detail::double_min) {
+            if (x < detail::sqrt_double_min) {
                 w = detail::j0_YP[7] / detail::j0_YQ[6];
                 w += M_2_PI * std::log(x);
                 return (w);
--- a/include/xsf/cephes/j0.h
+++ b/include/xsf/cephes/j0.h
@@ -105,8 +105,6 @@ namespace cephes {
 
     namespace detail {
 
-        inline constexpr double sqrt_double_min = 1.4916681462400413e-154;
-
         constexpr double j0_PP[7] = {
             7.96936729297347051624E-4, 8.28352392107440799803E-2, 1.23953371646414299388E0,  5.44725003058768775090E0,
             8.74716500199817011941E0,  5.30324038235394892183E0,  9.99999999999999997821E-1,
@@ -169,14 +167,13 @@ namespace cephes {
         }
 
         if (x <= 5.0) {
-            if (x < 1.0e-5) {
-                /* For x < sqrt_double_min (~1.49e-154), x*x underflows to zero and raises a
-                 * spurious floating-point exception. j0(x) = 1 - x^2/4 + ...
-                 * is 1.0 to machine precision when x*x underflows.
+            if (x < 3e-8) {
+                /* For x < 3e-8, 1 - x^2/4 rounds to 1.0 in double precision
+                 * (x^2/4 is below machine epsilon). Avoid computing x*x entirely.
                  */
-                if (x < detail::sqrt_double_min) {
-                    return (1.0);
-                }
+                return (1.0);
+            }
+            if (x < 1.0e-5) {
                 z = x * x;
                 return (1.0 - z / 4.0);
             }
@@ -220,12 +217,10 @@ namespace cephes {
                 set_error("y0", SF_ERROR_DOMAIN, NULL);
                 return std::numeric_limits<double>::quiet_NaN();
             }
-            /* When x < sqrt_double_min (~1.49e-154), x*x underflows to zero, which raises a
-             * spurious floating-point exception. Avoid the underflow by using
-             * the limiting form directly: for tiny x, j0(x) ≈ 1 and the
-             * rational approximation reduces to its constant term.
+            /* When x < 3e-8, x*x/4 is below machine epsilon for double precision.
+             * Avoid computing x*x entirely by using the limiting form directly.
              */
-            if (x < detail::sqrt_double_min) {
+            if (x < 3e-8) {
                 w = detail::j0_YP[7] / detail::j0_YQ[6];
                 w += M_2_PI * std::log(x);
                 return (w);
--- a/include/xsf/cephes/j0.h
+++ b/include/xsf/cephes/j0.h
@@ -105,6 +105,8 @@ namespace cephes {
 
     namespace detail {
 
+        inline constexpr double double_min = 1.4916681462400413e-154;
+
         constexpr double j0_PP[7] = {
             7.96936729297347051624E-4, 8.28352392107440799803E-2, 1.23953371646414299388E0,  5.44725003058768775090E0,
             8.74716500199817011941E0,  5.30324038235394892183E0,  9.99999999999999997821E-1,
@@ -168,11 +170,11 @@ namespace cephes {
 
         if (x <= 5.0) {
             if (x < 1.0e-5) {
-                /* For very small x, x*x underflows to zero and raises a
+                /* For x < double_min (~1.49e-154), x*x underflows to zero and raises a
                  * spurious floating-point exception. j0(x) = 1 - x^2/4 + ...
                  * is 1.0 to machine precision when x*x underflows.
                  */
-                if (x < 2e-154) {
+                if (x < detail::double_min) {
                     return (1.0);
                 }
                 z = x * x;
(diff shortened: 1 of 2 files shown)


include/xsf/cephes/j0.h
```diff
@@ -167,10 +167,18 @@ namespace cephes {
         }
 
         if (x <= 5.0) {
-            z = x * x;
             if (x < 1.0e-5) {
+                /* For very small x, x*x underflows to zero and raises a
+                 * spurious floating-point exception. j0(x) = 1 - x^2/4 + ...
+                 * is 1.0 to machine precision when x*x underflows.
+                 */
+                if (x < 2e-154) {
```
dschmitz89: Please add the sqrt(double_min) somewhere as a constexpr if it is not already available for example in cephes/const.h .

dschmitz89 (dismissed):
Thanks @trustdemons05 . Could you be more explicit about the AI usage here? 

The changes make sense but this needs some more work to be acceptable for merging. Most importantly we need some tests for the new extra branches. Let us know if you need help with those.

trustdemons05:
> Thanks @trustdemons05 . Could you be more explicit about the AI usage here?
> 
> The changes make sense but this needs some more work to be acceptable for merging. Most importantly we need some tests for the new extra branches. Let us know if you need help with those.

I used an AI coding assistant (opencode) to help draft parts of this fix, including the threshold constant extraction and test scaffolding. all code was reviewed, understood, and verified by me before submission.

include/xsf/cephes/j0.h
steppi: you can define it in this file as something like `inline constexpr double sqrt_double_min`

steppi:
> The changes make sense but this needs some more work to be acceptable for merging. Most importantly we need some tests for the new extra branches. Let us know if you need help with those.

These functions have a lot of test cases and coverage may already exist in the parquet files. We should check what the coverage report says.

include/xsf/cephes/j0.h
```diff
@@ -105,6 +105,8 @@ namespace cephes {
 
     namespace detail {
 
+        inline constexpr double double_min = 1.4916681462400413e-154;
```
steppi: Sorry, I had a typo in my original comment.
```suggestion
        inline constexpr double sqrt_double_min = 1.4916681462400413e-154;
```

trustdemons05: done renamed to sqrt_double_min

trustdemons05:
> These functions have a lot of test cases and coverage may already exist in the parquet files. We should check what the coverage report says.

Good point, I checked the scipy_special_tests parquet tables for cyl_bessel_j0 and cyl_bessel_y0 — they don't include values below ~1e-5, so the tiny-x branches (x < sqrt_double_min ≈ 1.49e-154) wouldn't be exercised by existing data. The hand-written tests in test_j0y0_tiny_x.cpp cover that gap.

tests/xsf_tests/test_j0y0_tiny_x.cpp
steppi, line 1: I think the tests here should be updated to actually check that underflow hasn't occurred.
You'll need to use `std::ftestexcept` from the `<cfenv>` header, and make sure to run `std::feclearexcept(FE_ALL_EXCEPT);` before each calculation. The example here, https://en.cppreference.com/cpp/numeric/fenv/fetestexcept) , should make it clear what to do.

trustdemons05: added std::feclearexcept(FE_ALL_EXCEPT) before each calculation and CHECK(std::fetestexcept(FE_UNDERFLOW) == 0) after

include/xsf/cephes/j0.h
```diff
@@ -167,10 +169,18 @@ namespace cephes {
...
 
         if (x <= 5.0) {
-            z = x * x;
             if (x < 1.0e-5) {
+                /* For x < sqrt_double_min (~1.49e-154), x*x underflows to zero and raises a
+                 * spurious floating-point exception. j0(x) = 1 - x^2/4 + ...
+                 * is 1.0 to machine precision when x*x underflows.
+                 */
+                if (x < detail::sqrt_double_min) {
+                    return (1.0);
```
dschmitz89: The culprit here is basically: what is the smallest x for which $1-x^2/4\ !=\ 1$ in double precision? I think that path will lead to another threshold. WDYT?

j-bowhay (commented):
It would be good to make the tests a little more DRY

include/xsf/cephes/j0.h
trustdemons05: The threshold would be 3e-8 , for any x between 1.5e-154 and 3e-8, 1 − x²/4 rounds back to 1.0. I think we can simplify by replacing both checks with a single if (x < 3e-8) return 1.0 and only compute 1 − x²/4 when x ≥ 3e-8.  If you like the solution , I can implement this exact fix.

trustdemons05:
> It would be good to make the tests a little more DRY

refactored with a CHECK_NO_UNDERFLOW helper macro

include/xsf/cephes/j0.h
dschmitz89: Sounds good 👍, thanks.

tests/xsf_tests/test_j0y0_tiny_x.cpp
```diff
@@ -0,0 +1,67 @@
+#include "../testing_utils.h"
+#include <cfenv>
+#include <xsf/cephes/j0.h>
+
+#define CHECK_NO_UNDERFLOW(expr)                                                           \
```
j-bowhay: Sorry I should have been clearer, I'd rather use the built-in test case generation or a simple for loop rather than a macro https://github.com/scipy/xsf/blob/daf8bb6ac485f34f7d6b212c3b8bdb5e60a54b77/tests/xsf_tests/test_ndtri_reparametrizations.cpp#L12

trustdemons05: oh thats my bad i should have used generate all along.

tests/xsf_tests/test_j0y0_tiny_x.cpp
```diff
@@ -0,0 +1,64 @@
...
+TEST_CASE("j0 tiny x underflow threshold", "[j0][y0][xsf_tests]") {
+    // Regression test for gh-25199: j0/y0 at tiny x should not raise spurious
+    // FP exceptions from x*x underflowing to zero.
+    //
+    // Reference values computed with mpmath with 1000 digits of precision:
+    //
+    // import mpmath as mp
+    // mp.mp.dps = 1000
+    // for x in [0.0, 1e-300, 1e-200, 1e-100, 1e-10, 1e-5]:
+    //     print(x, mp.nstr(mp.besselj(0, x), 18), mp.nstr(mp.bessely(0, x), 18))
```
dschmitz89, line 14: Do you not need to cast `x` to a multiprecision variable (sth. like mp.mpf) before using it in the mpmath calculation?

tests/xsf_tests/test_j0y0_tiny_x.cpp
```diff
@@ -0,0 +1,64 @@
...
+        CHECK(w < 0.0);
+        CHECK(std::fetestexcept(FE_UNDERFLOW) == 0);
+    }
+
+    SECTION("y0 at 2e-154 is finite and negative") {
+        std::feclearexcept(FE_ALL_EXCEPT);
+        auto y = xsf::cephes::y0(2e-154);
+        CHECK(std::isfinite(y));
+        CHECK(y < 0.0);
+        CHECK(std::fetestexcept(FE_UNDERFLOW) == 0);
```
dschmitz89: Can those two tests be combined into one? They essentially test the same thing.

include/xsf/cephes/j0.h
```diff
@@ -167,10 +167,17 @@ namespace cephes {
         }
 
         if (x <= 5.0) {
-            z = x * x;
+            if (x < 3e-8) {
+                /* For x < 3e-8, 1 - x^2/4 rounds to 1.0 in double precision
+                 * (x^2/4 is below machine epsilon). Avoid computing x*x entirely.
```
j-bowhay:
```suggestion
```

include/xsf/cephes/j0.h
```diff
@@ -210,6 +217,14 @@ namespace cephes {
                 set_error("y0", SF_ERROR_DOMAIN, NULL);
                 return std::numeric_limits<double>::quiet_NaN();
             }
+            /* When x < 3e-8, x*x/4 is below machine epsilon for double precision.
+             * Avoid computing x*x entirely by using the limiting form directly.
```
j-bowhay:
```suggestion
```

j-bowhay (commented):
Some of the comments are a bit overly verbose

j-bowhay:
Also please could you update the pr description to reflect the current implementation

tests/xsf_tests/test_j0y0_tiny_x.cpp
steppi: If you're just passing a variable directly into the function,  `mpmath` takes care of it.

See

https://github.com/mpmath/mpmath/blob/939efcdd88c455380ac12b1b688f9160d85c3bf1/mpmath/functions/bessel.py#L25

and

https://github.com/mpmath/mpmath/blob/939efcdd88c455380ac12b1b688f9160d85c3bf1/mpmath/ctx_mp_python.py#L823

You only need to worry about converting to an `mp.mpf` or `mp.mpc` before doing artithmetic operations with the variables. So `mp.besselj(0, 1/x)` is incorrect but `mp.besselj(0, x)` and `mp.besselj(0, mp.one/x)` are fine.

trustdemons05: I agree with steppi , Can i mark this as resolved?

tests/xsf_tests/test_j0y0_tiny_x.cpp
trustdemons05: yep , done thanks for pointing it out

trustdemons05:
> Also please could you update the pr description to reflect the current implementation

Done

trustdemons05:
> Also please could you update the pr description to reflect the current implementation

Should i also squash the commits?

j-bowhay:
> > Also please could you update the pr description to reflect the current implementation
> 
> Should i also squash the commits?

That's probably not necessary, we can squash merge

steppi (approved):
Thanks @trustdemons05. I think this is a paradigmatic example of a good AI-assisted contribution. Limited in scope. Solves a real problem in a principled way. Developer was responsive to maintainer feedback and demonstrated understanding of the changes. This is exactly how I thought this problem should actually be solved when I saw https://github.com/scipy/scipy/pull/25202.

Issue #163, opened 2026-05-24
MAINT: Headers should be reorganized to cleanly demarcate between cpu/cuda capable code and cpu-only code


steppi:
While working on  https://github.com/cupy/cupy/pull/9959, I had to do some partial reorganizing to get the Poisson Binomial gufuncs working in CuPy. This is become the Poisson Binom kernels live in `xsf/include/xsf/stats.h`, and this contained a mix of CPU+GPU code and CPU-only code. I think it's about time that we systematically organize the headers to make it clear which functions are supported on which platforms, and to ensure that NVRTC won't choke due to some CPU-only code being mixed into a header that is supposed to be GPU only.

Also, we need to make sure to only use relative includes within `xsf`. I saw that some things like `#include "xsf/config.h"` have snuck in recently. This will break on CuPy because `xsf` doesn't exist at the top level include path there, but falls under `include/cupy/xsf`. 

I propose a directory structure like

`xsf/include/xsf/...` for host + device capable code
`xsf/include/xsf/cpu/...` for cpu-only code (and functions will go in an `xsf::cpu` namespace.
`xsf/include/xsf/numpy.h` can be grandfathered in at the top level, and this stuff is in the `xsf::numpy` namespace anyway and is already clearly demarcated.

Rather than doing this all in one go, I want to reorganize in this fashion on a piece-meal basis, as things are touched. So as I work to complete the changes needed for https://github.com/cupy/cupy/pull/9959, I will only make this split for what is now `stats.h`. For any header which is intended to work on CPU/GPU I will make sure there are tests which confirm no CPU-only things have snuck in. I will also look into adding tests to make sure only relative imports are used.

@dschmitz89, @fbourgey I figured it would be best to post this to make sure we're all on the same page about this before I make a PR for the changes needed for https://github.com/cupy/cupy/pull/9959.

fbourgey:
Just to be clear, everything without `XSF_HOST_DEVICE` should be moved to `xsf/include/xsf/cpu/`? what to do with `xsf/include/xsf/amos/`, `xsf/include/xsf/cephes/`, `xsf/include/xsf/specfunc/`

steppi:
> Just to be clear, everything without `XSF_HOST_DEVICE` should be moved to `xsf/include/xsf/cpu/`? 

[...]

PR #164, opened 2026-05-25
ENH: Make changes needed to add poisson_binom gufuncs to CuPy

steppi:

This PR adds a templated function for creating `mdspan` views equivalent to a CuPy `CArray` view of a buffer. `CArray` is essentially CuPy's equivalent of `mdspan` from before `mdspan` ever became a thing. In order to add a gufunc-like `BatchwiseKernel` to CuPy here, https://github.com/cupy/cupy/pull/9959, I updated CArray to have a template argument for the core dimension, and added overloads for `operator[]` which allow indexing only the batch dimensions, and getting a `CArray` view of the corresponding core slice. The kernel from `xsf` is then handed this `CArray`, and `xsf`'s `as_mdspan` is used to essentially "turn it into" an `mdspan` for use in `xsf` kernels which expect `mdspan`s for arguments with core dimension greater than 0.

I've made this `as_mdspan` rely on the `libcu++` `mdspan` when compiling with CUDA and the vendored `kokkos` mdspan otherwise, despite this tool being GPU only. This is to enable support for ROCm/HIP, where the `kokkos` mdspan should work fine.

I've also moved the parts of `stats.h` which are cpu-only and cause NVRTC to choke into a separate `cpu/stats.h` as discussed here #163. I also had to make some other little changes to ensure that the GPU-supporting parts actually do work e.g. replacing `std::real(z)` with `z.real()`, adding `XSF_HOST_DEVICE` annotations, adding `std::min` and `std::max` to `config.h`, (while adding `std::min` and `std::max` I also took the trouble to fix our CUDA implementation of `std::clamp`, which I allowed to remain broken because it isn't actually used anywhere in `xsf` yet).

#### AI Generation Disclosure

I asked Google Gemini 3.5 flash for advice on naming for the directory/namespace for cpu-only things in `xsf`. I had Google Gemini 3.1 pro rewrite a second draft of `cupy.h`  which swaps  between the `libcu++` `mdspan` and the `kokkos` ione depending on platform.


include/xsf/cpu/stats.h
```diff
@@ -0,0 +1,143 @@
...
+        }
+        if (x <= 0) {
+            return 0.0;
+        }
+
+        double total = 0.0;
+        int k = 0;
+
+        while (true) {
+            double u = std::exp(gammaln(k + 0.5) - gammaln(k + 1.0)) / (std::pow(M_PI, 1.5) * std::sqrt(x));
```
dschmitz89, line 44: Can the gamma ratio not be expressed directly with the `poch` function?

steppi: Perhaps, but it's not relevant for this PR. The reason this shows up in the diff is because `cdf_cvm_inf`  depends on Bessel functions which aren't supported on GPU yet, so I've moved this into `xsf/cpu/stats.h` so that I can include `xsf/stats.h` in the preamble of a kernel in CuPy. There are similar reasons for all of the functions I've moved here.

steppi:
Marking as draft because Sebastian suggested moving `as_mdspan` to CuPy in https://github.com/cupy/cupy/pull/9959, and I want to make sure this doesn't get merged until I take it out here.

steppi:
I've removed `cupy.h`

steppi:
@gpinkert, this PR is the current blocker towards updating xsf in CuPy to get #156 in over there. Can you help review? Once we get this in, I can make the corresponding CuPy PR. Also, GPU CI will run eventually, it's just that a single runner is being shared between xsf and SciPy.

include/xsf/cpu/stats.h
```diff
@@ -0,0 +1,143 @@
...
+            return x;
+        }
+        if (x <= 0) {
+            return 0.0;
+        }
+
+        double total = 0.0;
+        int k = 0;
+
+        while (true) {
```
gpinkert, line 43: I think a do - while loop would be more appropriate here

include/xsf/cpu/stats.h
```diff
@@ -0,0 +1,143 @@
...
+
+    inline double smirnovci(int n, double x) { return cephes::smirnovci(n, x); }
+
+    inline double smirnovp(int n, double x) { return cephes::smirnovp(n, x); }
+
+    inline double cdf_cvm_inf(double x) {
+        // CDF of the Cramer-von Mises test statistic (infinite sample limit).
+        // Accurate for practical hypothesis testing but not expected to be accurate
+        // for large values of x, e.g. x > 4, when the cdf is very close to 1.
+        if (std::isnan(x)) {
```
gpinkert, line 33: I don't know much about the math, but do you not need to check for infinity?

include/xsf/cpu/stats.h
gpinkert: actually, if you are incrementing K, why would you not make this a for loop? Something like:

```c++
double z = 0.0f;
for (int k = 0; std::abs(z) >= 1e-7; ++k)
{
     // Body
}
```

include/xsf/cpu/stats.h
```diff
@@ -0,0 +1,143 @@
+#include "../bessel.h"
+#include "../cephes/kolmogorov.h"
+#include "../config.h"
+#include "../gamma.h"
+
+namespace xsf {
+namespace cpu {
+
+    inline double kolmogorov(double x) { return cephes::kolmogorov(x); }
```
gpinkert, line 9: Just curious, why not make these templates? Would get extra types for free, or you could explicitly disallow any type promotion.

include/xsf/cpu/stats.h
```diff
@@ -0,0 +1,143 @@
+#include "../bessel.h"
+#include "../cephes/kolmogorov.h"
+#include "../config.h"
+#include "../gamma.h"
+
+namespace xsf {
```
gpinkert, line 6: Not blocking but I think there is some guidelines somewhere that says to do 
```c++
namespace xsf::cpu
{
    // stuff
}
``` 

instead, but I don't care, just pointing out.

include/xsf/cpu/stats.h
```diff
@@ -0,0 +1,143 @@
...
+                break;
+            }
+            ++k;
+        }
+
+        return total;
+    }
+
+    // Helpers for psi1 computations
+    XSF_HOST_DEVICE inline double psi1_G(double y) {
```
gpinkert: also i am a little confused why the cpu namespaced functions are marked `XSF_HOST_DEVICE`. Wouldn't that imply that these are also device functions?

include/xsf/cpu/stats.h
```diff
@@ -0,0 +1,143 @@
...
+        // Helper implementing the G-type Bessel combination appearing in equation 1.10
+        // in Csörgő, S. and Faraway, J. (1996).
+        double z = y * y / 4.0;
+        double b = cyl_bessel_k(0.25, z) + cyl_bessel_k(0.75, z);
+        return std::exp(-z) * std::pow(y / 2.0, 1.5) * b / std::sqrt(M_PI);
```
gpinkert, line 66: Technically using `M_PI` isn't very portable since  you have to include a 
```c++
#define _USE_MATH_DEFINES
#include <cmath>
double x =M_PI;
```
if you want it to compile on MSVC. I don't know much about your compilation targets but IIRC there is a windows build so I thought that I would point that out.

include/xsf/cpu/stats.h
```diff
@@ -0,0 +1,143 @@
...
+        // in Csörgő, S. and Faraway, J. (1996).
+        double z = y * y / 4.0;
+        double b = 2.0 * cyl_bessel_k(0.25, z) + 3.0 * cyl_bessel_k(0.75, z) - cyl_bessel_k(1.25, z);
+        return std::exp(-z) * std::pow(y / 2.0, 2.5) * b / std::sqrt(M_PI);
+    }
+
+    XSF_HOST_DEVICE inline double psi1_term(int k, double x) {
+        // Compute the k-th term of the psi1 series expansion in equation 1.10
+        // of Csörgő, S. and Faraway, J. (1996).
+        double m = 2.0 * k + 1.0;
```
gpinkert, line 80: Just a comment that you could likely const most of these variables. Might result in better codegen, might not. up to you but not a blocker.

include/xsf/config.h
```diff
@@ -144,15 +147,28 @@ XSF_HOST_DEVICE inline double fmod(double x, double y) { return ::fmod(x, y); }
...
+
 template <typename T>
 XSF_HOST_DEVICE void swap(T &a, T &b) {
     cuda::std::swap(a, b);
 }
 
-// Reimplement std::clamp until it's available in CuPy
+// Reimplement std::min, std::max, std::clamp until they are available in CuPy
+template <typename T>
+XSF_HOST_DEVICE constexpr const T &min(const T &a, const T &b) {
```
gpinkert, line 160: I notice that you have some cuda stl headers included in this file. What is stopping us from `#include <cuda/std/algorithm>` so that we can get min/max/clamp for free? I probably just don't understand how XSF work in cupy

include/xsf/stats.h
```diff
@@ -1,181 +1,58 @@
...
+#include "cephes/ndtr.h"
+#include "cephes/ndtri.h"
+#include "cephes/owens_t.h"
+#include "cephes/pdtr.h"
+#include "cephes/tukey.h"
+#include "config.h"
+#include "erf.h"
+#include "gamma.h"
 
 namespace xsf {
```
gpinkert, line 21: Should we consider putting this in a `xsf::device` namespace, to match the cpu namespace?

gpinkert (commented):
I could approve as well but it feels a little self important of me, I don't know the codebase or the goals well enough one way or another

include/xsf/cpu/stats.h
steppi: Yeah, this is something that could be cleaned up now. When Irwin and I started what became`xsf`, SciPy was still on C++14 so we couldn't use the nice nested namespace syntax. What we have here is consistent with other nested namespaces in the codebase. I think the cleanup should be for all nested namespaces at once in a single PR.

include/xsf/cpu/stats.h
steppi: For functions that don't have specific single precision implementations, SciPy has always cast `float`s to `double`s and then cast the result back to `float`. For the majority of special functions in `xsf` you would only get the illusion of getting extra types for free. Many of the implementations are precision specific. For instance, minimax polynomial approximations have to be tuned to a specific precision and you can get significantly inaccurate results using the `double` polynomial for `float`s due to accumulated rounding error.

include/xsf/cpu/stats.h
steppi: @fbourgey translated this from an existing Python or Pythran implementation (I forget which). When we do translations like that, often we'll make them as mechanical as possible to make the diff easier to review, with the thought that things could always be improved in a follow-up. This isn't in-scope for this PR though, which is only about separating out some host-only code into separate headers to get things working in CuPy again.

include/xsf/cpu/stats.h
steppi: Good catch. I missed that these got added back in when I resolved merge conflicts recently.

include/xsf/cpu/stats.h
steppi: We define the math constants in `config.h` on platforms where they don't exist. This has been a longstanding practice in SciPy

https://github.com/scipy/xsf/blob/ce37b1c30650d137bb66288bd477e441c2278ed9/include/xsf/config.h#L24-L26

include/xsf/cpu/stats.h
steppi: Yeah, I think this is a good idea in general, but it's not in scope for this PR.

include/xsf/config.h
steppi: `max`, `min`, and `clamp` are not part of the public interface of `libcudacxx` yet. There are some private implementations `libcudacxx` uses internally though, and I plan to ask them if they could make them public.

include/xsf/stats.h
steppi: I thought about that, but then almost all of `xsf` would end up with the `device` namespace. Having `cpu` only code is only a temporary thing, and the goal is to eventually have only device capable stuff in `xsf`. I am working on a project roadmap where I will lay out my plan for `xsf` to be a library only for single source cpu/gpu implementations of ufunc and gufunc kernels.

steppi (commented):
Thanks @gpinkert for taking a look, and also for taking the chance to familiarize yourself with the codebase a little and ask questions about things that look strange even if they aren't in-scope for this PR. I hope I've done a decent job of answering the questions. I will remove the stray `XSF_HOST_DEVICE`'s that made their way into the `cpu/stats.h` header when resolving merge conflicts.

include/xsf/cpu/stats.h
steppi: Huh, I'm not sure why my response didn't get posted earlier. Since `x` only appears in the denominators in the terms below, this will produce the correct result without checking for infinity. It might still be desirable to check though, but see my other comment about this being a mechanical translation of an existing implementation.

gpinkert (approved):

steppi:
Thanks @gpinkert! I'll go ahead and merge now and make a CuPy PR later today.

PR #169, opened 2026-06-01
MAINT: Use correct name for bivariate normal survival function

fbourgey:
#### Reference issue
None

While working on translating the MATLAB code for the trivariate normal distribution on https://github.com/scipy/xsf/issues/98, I noticed that the name `bivariate_normal_cdf` in https://github.com/scipy/xsf/pull/112 is incorrect. This is really the survival function as it computes $P(X > dh, Y>dk)$.

I have renamed everything to `bivariate_normal_sf` and updated the documentation. 

If we agree with the changes, I will follow up with a SciPy PR.

#### AI Generation Disclosure

No AI.

### DIFF
--- a/include/xsf/stats.h
+++ b/include/xsf/stats.h
@@ -269,8 +269,8 @@ inline std::complex<float> log_ndtr(std::complex<float> z) {
 
 inline double nbdtr(int k, int n, double p) { return cephes::nbdtr(k, n, p); }
 
-XSF_HOST_DEVICE inline bool bivariate_normal_cdf_boundary(double dh, double dk, double r, double &p) {
-    // Handles degenerate cases for bivariate_normal_cdf (infinite arguments or zero correlation).
+XSF_HOST_DEVICE inline bool bivariate_normal_sf_boundary(double dh, double dk, double r, double &p) {
+    // Handles degenerate cases for bivariate_normal_sf (infinite arguments or zero correlation).
     // Returns true and sets p if a boundary case applies, false otherwise.
 
     const double math_inf = std::numeric_limits<double>::infinity();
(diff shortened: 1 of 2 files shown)


fbourgey:
@dschmitz89 WDYT?

dschmitz89:
I am confused. Do we not use this for the CDF in SciPy? If it was instead the survival function, would we or some user not have noticed?

fbourgey:
This was a direct port of https://github.com/scipy/scipy/blob/a4463edad21e858fb7556b4fd121463da8da2e33/scipy/stats/_stats_pythran.py#L224-L355

Now, it is only used here https://github.com/scipy/scipy/blob/main/scipy/stats/_qmvnt.py

The tests in `xsf_tests/test_bivariate_normal_cdf.cpp` were testing the old `_bvnu` implementation, and some symmetry and complementary. I should have been more careful, and test it with something different than SciPy. Apologies for this.

The following script (run on scipy/main) shows that this is really the SF.

```python
In [6]: import numpy as np
   ...: from scipy.special._ufuncs import _bivariate_normal_cdf as _bvnu
   ...: from scipy.stats import multivariate_normal
   ...: 
   ...: rng = np.random.default_rng(12345)
   ...: dhs = rng.uniform(-5, 5, 5)
   ...: dks = rng.uniform(-5, 5, 5)
   ...: rs = rng.uniform(-1.0, 1.0, 5)
   ...: for dh, dk, r in zip(dhs, dks, rs):
   ...:     out = _bvnu(dh, dk, r)
   ...:     print(f"bvnu({dh:.2f}, {dk:.2f}, {r}) = {out:.15e}")
   ...: 
   ...:     cdf1 = multivariate_normal.cdf([dh, dk], mean=[0, 0], cov=[[1, r], [r, 1]])
   ...:     print(cdf1)
   ...: 
   ...:     cdf2 = multivariate_normal.cdf([-dh, -dk], mean=[0, 0], cov=[[1, r], [r, 1]])
   ...:     print(cdf2)
   ...: 
   ...:     assert np.allclose(out, cdf2)
   ...: 
   ...:     print()
bvnu(-2.73, -1.67, -0.503508570740858) = 9.495253611653549e-01
3.9839167553790134e-07
0.9495253611653548

bvnu(-1.83, 0.98, 0.8977623036666365) = 1.627821790862717e-01
0.03344468815266366
0.16278217908627177

bvnu(2.97, -3.13, 0.3344749062007448) = 1.471374824124958e-03
0.0008661494764053953
0.0014713748241249745

bvnu(1.76, 1.73, -0.8082041288117758) = 3.800814910790886e-10
0.9189779841115192
3.8008152181134847e-10

bvnu(-1.09, 4.42, -0.11632066766437443) = 3.537834819241452e-06
0.13809655971558984
3.5378348191561493e-06
```

fbourgey:
The rename was done following your comment https://github.com/scipy/xsf/pull/112#pullrequestreview-4165634292 that we should use something more explicit than `_bvnu`. Btw, it probably means "bivariate normal upper"! This is the upper-orthant probability or survival function.

dschmitz89:
@fbourgey just to clarify: the xsf function that is renamed here is not used in scipy's multivariate normal CDF yet actually, thats why it went unnoticed?

fbourgey:
Looking at the SciPy code, it appears that [multivariate_normal.cdf](https://github.com/scipy/scipy/blob/main/scipy/stats/_qmvnt.py#L183) goes through [_bvn](https://github.com/scipy/scipy/blob/main/scipy/stats/_qmvnt.py#L462-L475) in the 2D case, and `_bvn` calls `_bvnu`. This seems to be the only place. 
But _bvnu itself seems to be the upper-tail helper; _bvn turns that into the rectangle probability.

dschmitz89:
> Looking at the SciPy code, it appears that [multivariate_normal.cdf](https://github.com/scipy/scipy/blob/main/scipy/stats/_qmvnt.py#L183) goes through [_bvn](https://github.com/scipy/scipy/blob/main/scipy/stats/_qmvnt.py#L462-L475) in the 2D case, and `_bvn` calls `_bvnu`. This seems to be the only place. But _bvnu itself seems to be the upper-tail helper; _bvn turns that into the rectangle probability.

Allright. Then lets try to make the naming clearer on the SciPy side as well in a followup, WDYT?

Is there sth. like `_bvnl`/is it needed for the CDF ?

fbourgey:
IIUC, you are suggesting a _bvnl (bivariate normal lower) helper. I do not think we need it as `_bvn` appears to compute P(a0 <= X <= b0, a1 <= Y <= b1). However, documentation for `_bvn` is lacking so I could strengthen it.

Issue #170, opened 2026-06-04
ENH: Move `scipy.special` kernels to XSF


fbourgey:
Several kernels still live in `scipy/special/` and look suitable for migration to XSF.

  - [x] [_convex_analysis.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_convex_analysis.pxd) #171 
  - [x] [_agm.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_agm.pxd) #172 
  - [x] [_boxcox.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_boxcox.pxd) #178
  - [x] [_ndtri_exp.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_ndtri_exp.pxd) #180
  - [x] [_hypergeometric.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_hypergeometric.pxd) #183 
  - [x] [_spence.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_spence.pxd) #191
  - [x] [_hyp0f1.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_hyp0f1.pxd) #197
  - [x] [_cosine.c](https://github.com/scipy/scipy/blob/main/scipy/special/_cosine.c) #198 
  - [x] [stirling2.h](https://github.com/scipy/scipy/blob/main/scipy/special/stirling2.h) #213 
  - [x] [gen_harmonic.h](https://github.com/scipy/scipy/blob/main/scipy/special/gen_harmonic.h). #210 
  - [ ] [orthogonal_eval.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/orthogonal_eval.pxd).
      - https://github.com/scipy/xsf/pull/209
  - [x] [wright.c](https://github.com/scipy/scipy/blob/main/scipy/special/wright.cc) #199
  - [ ] [ellint_carlson_cpp_lite.cxx](https://github.com/scipy/scipy/blob/main/scipy/special/ellint_carlson_wrap.cxx) https://github.com/scipy/xsf/pull/271

Suggested approach:

  1. Add header-only implementations and focused tests in XSF.
  2. Update SciPy to route the corresponding ufunc kernels through XSF and remove old code in a follow-up SciPy PR.

dschmitz89:
[...]

PR #171, opened 2026-06-04
ENH: add `convex_analysis` functions to XSF

fbourgey:
#### Reference issue

- Port [_convex_analysis.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_convex_analysis.pxd) to xsf
- Tests are direct translation of SciPy tests except for `pseudo_huber` which uses random values

#### AI Generation Disclosure

Tests were translated by Codex. I manually added the links and few things for pseudo_huber. The rest is me. I've kept the comments.

### DIFF
--- a/include/xsf/convex_analysis.h
+++ b/include/xsf/convex_analysis.h
@@ -1,8 +1,5 @@
 #pragma once
 
-#include <cmath>
-#include <limits>
-
 #include "config.h"
 
 namespace xsf {
--- a/include/xsf/convex_analysis.h
+++ b/include/xsf/convex_analysis.h
@@ -1,0 +1,91 @@
+#pragma once
+
+#include <cmath>
+#include <limits>
+
+#include "config.h"
+
+namespace xsf {
+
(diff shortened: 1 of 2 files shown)


fbourgey:
@steppi I see that all functions are available in CuPy.

For example`kl_div` exists [here](https://docs.cupy.dev/en/stable/reference/generated/cupyx.scipy.special.kl_div.html). Should tests be added then?

steppi (approved):
Looks good!

fbourgey:
> @steppi I see that all functions are available in CuPy.
> 
> For example`kl_div` exists [here](https://docs.cupy.dev/en/stable/reference/generated/cupyx.scipy.special.kl_div.html). Should tests be added then?

@steppi so no CuPy tests?

steppi:
> @steppi so no CuPy tests?

No, not needed yet. CuPy currently has its own duplicate implementation. The plan is to swap out all of their implementations for xsf implementations, but not all at once. We can add the CuPy tests as things are switched into CuPy.

PR #172, opened 2026-06-04
ENH: add `agm` to XSF

fbourgey:
#### Reference issue

- Add arithmetic-geometric mean `agm` to XSF.

#### AI Generation Disclosure

Tests were translated by Codex. I manually added the links. The rest is mine. I've kept the SciPy comments.

### DIFF
--- a/include/xsf/agm.h
+++ b/include/xsf/agm.h
@@ -1,0 +1,82 @@
+#pragma once
+
+#include <cmath>
+#include <limits>
+
+#include "cephes/ellpk.h"
+#include "config.h"
+
+namespace xsf {
+
+namespace detail {
+
+    XSF_HOST_DEVICE inline double agm_iter(double a, double b) {
+        // Arithmetic-geometric mean, iterative implementation
+        // a and b must be positive (not zero, not nan).
+        double amean = 0.5 * a + 0.5 * b;
+        int count = 20;
+        while (count > 0 && (amean != a && amean != b)) {
+            double gmean = std::sqrt(a) * std::sqrt(b);
+            a = amean;
+            b = gmean;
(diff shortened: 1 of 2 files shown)


include/xsf/agm.h
```diff
@@ -0,0 +1,82 @@
+#pragma once
+
+#include <cmath>
```
steppi: We can't include these stdlib headers directly like this. Everything from the standard library needs to be included through `config.h` so we can use `cuda::std` for `std` when compiling for CUDA.

fbourgey: Thanks @steppi, I forgot about this. Not so easy to see those without GPU tests or CI failing.

steppi: Yeah, that's understandable. I think we can set up an automated check that the only includes outside of `config.h` and `numpy.h` are relative includes from within xsf itself.

include/xsf/agm.h
```diff
@@ -0,0 +1,79 @@
...
+        a = -a;
+        b = -b;
+    }
+
+    // At this point, a and b are both positive and not nan.
+
+    if ((invsqrthalfmax < a && a < sqrthalfmax) && (invsqrthalfmax < b && b < sqrthalfmax)) {
+        double e = 4 * a * b / ((a + b) * (a + b));
+        return sgn * (M_PI / 4) * (a + b) / cephes::ellpk(e);
+    }
```
steppi, line 69: That's a nice thing about programming in a compiled language. The compiler has our back here and will generate assembly that only computes `a + b` once. The way you have it is easier to read imo than if we created a temporary variable for `a + b`.

steppi (approved):
Looks good. Thanks @fbourgey!

PR #41, opened 2025-06-16
DEV/CI: parallel workers task args

lucascolley:
closes gh-34

PR #173, opened 2026-06-05
DEV: improvements to Pixi manifest

lucascolley:
I was looking into reviving gh-41 and noticed quite a few other minor improvements I would like to make first.

@fbourgey would you be willing to review? Probably easiest to review each commit individually.

There are some more changes I have in mind but they could wait for a follow-up.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -9,7 +9,6 @@ name = "xsf"
 platforms = ["win-64", "linux-64", "linux-aarch64", "osx-64", "osx-arm64"]
 version = "0.2.2"
 
-## Build
 
 [feature.build.dependencies]
 cmake = ">=3.30.5,<4"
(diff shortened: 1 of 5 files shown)


pixi.toml
```diff
@@ -131,34 +147,34 @@ configure-coverage.cmd = [
...
-build-tests-ci.description = "Build release test binaries for CI."
-# Run tests
-tests-ci.cmd = ["ctest", "--output-on-failure", "--test-dir", "build/tests", "-j3"]
-tests-ci.depends-on = ["build-tests-ci"]
-tests-ci.cwd = "."
-tests-ci.description = "Run release CTest suite for CI."
-# Coverage
-coverage.cmd = [
+[feature.tests-ci.tasks.build-tests-ci]
+description = "Build release test binaries for CI (configure for coverage)."
```
fbourgey: Do we mean?


```suggestion
description = "Build release test binaries for CI with coverage enabled."
```

pixi.toml
```diff
@@ -8,8 +8,29 @@ description = "Special function implementations."
 name = "xsf"
 platforms = ["win-64", "linux-64", "linux-aarch64", "osx-64", "osx-arm64"]
 version = "0.2.2"
+requires-pixi = ">=0.68.0"
```
fbourgey, line 11: Was wondering why we have `pixi-version: v0.70.1` elsewhere?

fbourgey (commented):
Thanks Lucas. Everything LGTM. One last thing: would it preferable to have the `feature.build.tasks.*` in alphabetical order? I see `[feature.build.tasks.configure]` before `[feature.build.tasks.build-only]`. Running `pixi task list` gives those in alphabetical order.

pixi.toml
lucascolley: v0.70.1 is the latest version. v0.68.0 is the oldest version that I think will still work with the current lock file format

lucascolley:
> Thanks Lucas. Everything LGTM. One last thing: would it preferable to have the `feature.build.tasks.*` in alphabetical order? I see `[feature.build.tasks.configure]` before `[feature.build.tasks.build-only]`. Running `pixi task list` gives those in alphabetical order.

I figured that since the `build`  task runs `configure` -> `build-only` in order, it makes sense for `configure` to appear first.

fbourgey (approved):

fbourgey:
Thanks!

PR #175, opened 2026-06-08
DEV: assorted Pixi workspace improvements

lucascolley:
@fbourgey another one! I think after this PR I can rebase gh-41. Probably easiest to review commits individually :)

depends on https://github.com/scipy/xsref/pull/17

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -34,10 +34,10 @@ solve-group = "default"
 
 [feature.build.dependencies]
 cmake = "*"
-make = ">=4.4.1,<5"
+make = "*"
 
 [feature.build.target.unix.dependencies]
-cxx-compiler = ">=1.8.0,<2"
+cxx-compiler = "*"
 
 [feature.build.target.win-64.dependencies]
 vs2026_win-64 = "*"
@@ -68,8 +68,8 @@ depends-on = ["configure", "build-only"]
 
 
 [feature.tests.dependencies]
-catch2 = ">=3.8.0,<4"
-libarrow-all = ">=19.0.1,<20"
+catch2 = "*"
+libarrow-all = "*"
 
 [feature.tests.tasks.clean-xsref]
 description = "Remove the local xsref."
@@ -141,7 +141,7 @@ cmd = "git ls-files '*.cpp' '*.h' | xargs clang-format --style=file --Werror --d
 
 
 [feature.coverage.dependencies]
-lcov = ">=1.16,<2"
+lcov = "*"
 
 [feature.coverage.tasks.configure-coverage]
 description = "Configure CMake for coverage instrumentation and tests."
@@ -167,7 +167,7 @@ cmd = ["open", "index.html"]
 
 
 [feature.tests-ci.dependencies]
-ccache = ">=4.11.2,<5"
+ccache = "*"
 
 # TODO: use a task arg for parallelism https://github.com/prefix-dev/pixi/pull/3433
 [feature.tests-ci.tasks.build-tests-ci]
@@ -218,10 +218,10 @@ cmd = [
 platforms = ["linux-64"]
 
 [feature.cupy-tests.dependencies]
-python = ">=3.12.0,<3.13"
+python = "*"
 pip = "*"
 setuptools = "*"
-cupy = ">=14.0.0"
+cupy = "*"
 pytest = "*"
 pytest-forked = "*"
 
--- a/pixi.toml
+++ b/pixi.toml
@@ -9,6 +9,7 @@ name = "xsf"
 platforms = ["win-64", "linux-64", "linux-aarch64", "osx-64", "osx-arm64"]
 version = "0.2.2"
 requires-pixi = ">=0.68.0"
+preview = ["pixi-build"]
 
 
 # environments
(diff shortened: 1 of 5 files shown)


fbourgey:
GPU tests failed.

steppi:
> GPU tests failed.

It looks like SciPy got brought in as a dependency for the CuPy tests despite only being an optional dependency of CuPy. I'm not sure why.

Update: Oh, it's because `xsref` is currently being installed as a package and it depends on an old SciPy. `xsref` doesn't actually need to depend on SciPy, that was supposed to be a temporary expedient. I'll remove the SciPy dependency tomorrrow.

lucascolley:
> Update: Oh, it's because `xsref` is currently being installed as a package and it depends on an old SciPy. `xsref` doesn't actually need to depend on SciPy, that was supposed to be a temporary expedient. I'll remove the SciPy dependency tomorrrow.

cool, let's return to this PR after that and https://github.com/scipy/xsref/pull/15

steppi:
Oh sorry, I guess this isn't just going to work out of the box, because the `pixi.lock` file needs to be updated.

pixi.toml
```diff
@@ -209,37 +213,11 @@ cmd = [
...
-fi
-'
-"""
-
-[feature.cupy-tests.tasks.install-xsref-test-cupy]
-description = "Install the local xsref package needed by CuPy tests."
-depends-on = ["clone-xsref-test-cupy"]
-cwd = "xsref"
-cmd = "pip install ."
+xsref = { git = "https://github.com/scipy/xsref", rev = "9504d5f270fa2e2aaa6006455019db88839aad50" }
```
lucascolley: @steppi hopefully you agree that this is an awesome simplification :)

steppi: Oh neat. This avoids reinstalling if `xsref` is already installed at this commit? I actually want to make it stop deleting and recloning when `xsref` is already available for all cases, not just the CuPy tests. Can you do that in this PR?

lucascolley: > Oh neat. This avoids reinstalling if `xsref` is already installed at this commit?

Yes!

> I actually want to make it stop deleting and recloning when `xsref` is already available for all cases, not just the CuPy tests. Can you do that in this PR?

Let me take a look

lucascolley: ah, the `tables` directory is not shipped in `xsref` wheels, so this doesn't actually work right now. Can we change it so that the `tables` directory is included @steppi ?

steppi: It sounds crazy, but if we're just installing it off github, that does seems like it should actually be fine.

lucascolley: okay, let's see if I managed to get this right

lucascolley:
could you check that the CI logs look to be picking up the xsref tables correctly @steppi ?

steppi:
> could you check that the CI logs look to be picking up the xsref tables correctly @steppi ?

Everything looks fine with the tables. The Windows failure is unrelated and is something that we had run into before and fixed. I'm not sure what retriggered it.

include/xsf/bessel.h
```diff
@@ -625,7 +625,8 @@ inline std::complex<double> cyl_bessel_je(double v, std::complex<double> z) {
 }
 
 inline std::complex<float> cyl_bessel_je(float v, std::complex<float> x) {
-    return static_cast<std::complex<float>>(cyl_bessel_je(static_cast<double>(v), static_cast<std::complex<double>>(x))
```
steppi, line 628: It's kind of annoying that `clang-format` changes its mind on how to format things almost every time we update `clang`. Can we keep it pinned to the previous version in this PR, and put clang updates in separate PRs, doing the `.git-blame-ignore-revs` thing for the style change commits?

lucascolley: Yep sounds good

tests/CMakeLists.txt
lucascolley, line 1: this has been generalised from requiring the `xsref` source to be cloned into the root directory, to having `xsref` installed in a Python virtual environment. Requires https://github.com/scipy/xsref/pull/17 to merge first.

pixi.toml
```diff
@@ -23,14 +24,22 @@ solve-group = "default"
 
 [environments.lint]
 features = ["clang-format"]
+no-default-feature = true
```
lucascolley, line 27: `xsref` is added to the default feature, but we don't need `xsref` (or `python`) for the `lint` env

pixi.toml
```diff
@@ -23,14 +24,22 @@ solve-group = "default"
...
 features = ["cupy-tests"]
 solve-group = "default"
 
 
-# features
+# default feature
+
+[dependencies]
+python = "*"
+xsref = { git = "https://github.com/lucascolley/xsref", rev = "a739aefa227c0a07f730dd8a1d1c8234ab48c815" }
```
lucascolley: TODO: point to `scipy/xsref`

include/xsf/bessel.h
lucascolley: this can be saved for a future PR which runs `pixi update`.

lucascolley:
this PR is now ready from my side! Reminder to review commits individually.

steppi (approved):
Looks great! Thanks @lucascolley, this is a major improvement

PR #178, opened 2026-06-08
ENH: Add `boxcox` to XSF

fbourgey:
#### Reference issue

- C++ translation of [scipy/special/_boxcox.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_boxcox.pxd) to XSF
- Add `boxcox`, `boxcox_1p, `boxcox_inv`, `boxcox_inv1p` to XSF

#### AI Generation Disclosure

I asked Codex to translate the tests from SciPy. I then quickly checked.

### DIFF
--- a/include/xsf/boxcox.h
+++ b/include/xsf/boxcox.h
@@ -1,0 +1,78 @@
+#pragma once
+
+#include "config.h"
+
+namespace xsf {
+
+XSF_HOST_DEVICE inline double boxcox(double x, double lmbda) {
+    /* If lmbda << 1 and log(x) < 1.0, the lmbda*log(x) product can lose
+     * precision, furthermore, expm1(x) == x for x < eps.
+     * For doubles, the range of log is -744.44 to +709.78, with eps being
+     * the smallest value produced.  This range means that we will have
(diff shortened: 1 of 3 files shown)


include/xsf/boxcox.h
```diff
@@ -0,0 +1,78 @@
+#pragma once
```
steppi, line 3: I like to add little comments at the top like this to credit the original authors.

```suggestion
// Translated from Cython into C++ by xsf developers in 2026.
// Original implementation by Warren Weckesser.
#pragma once
```

I missed that https://github.com/scipy/xsf/pull/171 should have such a comment too. The original author of that one is `argriffing`.

fbourgey: Sounds good. I added those.

dschmitz89 (approved):
LGTM.

dschmitz89:
Merging here. I am wondering if for such simple kernels we could use actual C++ templates instead of the brute force casting we do at the moment instead.

fbourgey:
> Merging here. I am wondering if for such simple kernels we could use actual C++ templates instead of the brute force casting we do at the moment instead.

I think so yes. There are some templated kernels already in XSF. 

@steppi

PR #179, opened 2026-06-09
BUG: handle huge complex inputs in `gamma` and `rgamma`

fbourgey:
#### Reference issue
None

- Patch for CI failing on `ubuntu-24.04-arm`. The error seemed to occur because of an overflowed `loggamma(z)` intermediate.

- Add guardrails for `gamma` and `rgamma` preventing those.
 
- Add corresponding tests. 

#### AI Generation Disclosure

Worked with Codex on this.

### DIFF
--- a/include/xsf/gamma.h
+++ b/include/xsf/gamma.h
@@ -57,7 +57,25 @@ XSF_HOST_DEVICE inline std::complex<double> gamma(std::complex<double> z) {
         set_error("gamma", SF_ERROR_SINGULAR, NULL);
         return {std::numeric_limits<double>::quiet_NaN(), std::numeric_limits<double>::quiet_NaN()};
     }
-    return std::exp(loggamma(z));
+
+    if (z.real() <= -std::ldexp(1.0, std::numeric_limits<double>::digits)) {
+        return {0.0, std::copysign(0.0, z.imag())};
+    }
+
+    std::complex<double> lg = loggamma(z);
+    if (lg.real() == -std::numeric_limits<double>::infinity()) {
+        return {0.0, std::copysign(0.0, z.imag())};
+    }
+    const double max = std::numeric_limits<double>::max();
+    if (lg.real() > std::log(max) && z.imag() == 0.0) {
+        return {std::numeric_limits<double>::infinity(), std::copysign(0.0, z.imag())};
+    }
+    if (lg.real() > std::log(max) && z.real() > std::sqrt(max) && std::abs(z.imag()) > std::sqrt(max)) {
+        return {
+            -std::numeric_limits<double>::infinity(), std::copysign(std::numeric_limits<double>::infinity(), z.imag())
+        };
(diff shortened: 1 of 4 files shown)


include/xsf/gamma.h
```diff
@@ -57,7 +57,25 @@ XSF_HOST_DEVICE inline std::complex<double> gamma(std::complex<double> z) {
         set_error("gamma", SF_ERROR_SINGULAR, NULL);
         return {std::numeric_limits<double>::quiet_NaN(), std::numeric_limits<double>::quiet_NaN()};
     }
-    return std::exp(loggamma(z));
+
+    if (z.real() <= -std::ldexp(1.0, std::numeric_limits<double>::digits)) {
+        return {0.0, std::copysign(0.0, z.imag())};
+    }
```
steppi, line 63: I see, so the idea is that if `z.real()` is less than the value at which all `double`s are even integers, we know not just that there was underflow, but also exactly what the signs of zero need to be because we know the parity of the real part of `z`.  I guess `loggamma(z)` is experiencing premature overflow in the `z.real() <= -2**53` regime, and this let's us get the correct signs of zero without going through `loggamma`. Very cool. It could use a comment explaining what's going on I think.

include/xsf/gamma.h
```diff
@@ -57,7 +57,25 @@ XSF_HOST_DEVICE inline std::complex<double> gamma(std::complex<double> z) {
...
+
+    std::complex<double> lg = loggamma(z);
+    if (lg.real() == -std::numeric_limits<double>::infinity()) {
+        return {0.0, std::copysign(0.0, z.imag())};
+    }
```
steppi, line 68: This should be able to handle the case for `z.real()` a very large magnitude negative int, but it seems reasonable to me to special case that and use this for `z` not so far from the origin but very close to a pole.

include/xsf/gamma.h
```diff
@@ -57,7 +57,25 @@ XSF_HOST_DEVICE inline std::complex<double> gamma(std::complex<double> z) {
...
+    const double max = std::numeric_limits<double>::max();
+    if (lg.real() > std::log(max) && z.imag() == 0.0) {
+        return {std::numeric_limits<double>::infinity(), std::copysign(0.0, z.imag())};
+    }
```
steppi, line 72: I see, so I guess the `exp` on ARM doesn't correctly handle this case and the case below. The spec actually doesn't say anything about what to return here, https://en.cppreference.com/cpp/numeric/complex/exp, so it seems good to separate it out anyway.

steppi (approved):
Thanks @fbourgey! Much appreciated. Everything looks good. Some explanatory comments would be good but can be done in a follow-up. It's nice to see CI green!

fbourgey:
Thanks @steppi. Like you said, it feels good to see CI finally turning green! I'll follow up with a PR adding comments.

PR #180, opened 2026-06-10
ENH: Add `ndtri_exp` to XSF

fbourgey:
#### Reference issue

- C++ translation of [scipy/special/_ndtri_exp.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_ndtri_exp.pxd) to XSF

#### AI Generation Disclosure

### DIFF
--- a/include/xsf/ndtri_exp.h
+++ b/include/xsf/ndtri_exp.h
@@ -1,0 +1,150 @@
+// Implementation of the inverse of the logarithm of the CDF of the standard
+// normal distribution.
+
+// Copyright: Albert Steppi
+
+// Distributed under the same license as SciPy
+
+// Implementation Overview
+
+// The inverse of the CDF of the standard normal distribution is available
+// in scipy through the Cephes Math Library where it is called ndtri.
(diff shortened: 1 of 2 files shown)


fbourgey:
- @steppi I created a separate header file as `stats.h` is quite large already. Let me know if we should put `ndtri_exp` in there too.

- For the tests, I used `linspace`. Should we use a random generator instead to match the SciPy tests?

steppi:
> * @steppi I created a separate header file as `stats.h` is quite large already. Let me know if we should put `ndtri_exp` in there too.

That sounds reasonable. I think `stats.h` should probably be broken up too. My original plan was to make the headers as granular as possible and to only run tests that depend on headers that have changed.

> * For the tests, I used `linspace`. Should we use a random generator instead to match the SciPy tests?

`linspace` is fine for now. After I get a chance to improve the infra for it, I'll add table based tests for the old test cases.

steppi (approved):
Thanks @fbourgey. This looks good. `ndtri_exp` was my first contribution to SciPy. It's been around 5 years. I don't think I could have imagined how things have turned out back then.

PR #187, opened 2026-06-12
ENH: add float overloads

fbourgey:
#### Reference issue
None

- Add float overloads. This is needed to clean up `scipy/special/functions.json`
- Added `XSF_HOST_DEVICE` in cephes/round.h following @steppi 's suggestion 
- Update float overload definition for `tukeylambdacdf`

#### AI Generation Disclosure

I started manually and then asked Codex to update the rest. I then checked and added a few more it missed.

### DIFF
--- a/include/xsf/specfun.h
+++ b/include/xsf/specfun.h
@@ -110,4 +110,8 @@ inline double pmv(double m, double v, double x) {
     return out;
 }
 
+inline float pmv(float m, float v, float x) {
+    return static_cast<float>(pmv(static_cast<double>(m), static_cast<double>(v), static_cast<double>(x)));
+}
+
 } // namespace xsf
--- a/include/xsf/cephes/erfinv.h
+++ b/include/xsf/cephes/erfinv.h
@@ -72,5 +72,7 @@ namespace cephes {
         }
     }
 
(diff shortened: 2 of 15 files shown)


include/xsf/specfun.h
```diff
@@ -110,4 +110,8 @@ inline double pmv(double m, double v, double x) {
     return out;
 }
 
+inline float pmv(float m, float v, float x) {
```
fbourgey, line 113: Following https://github.com/scipy/xsf/issues/163#issuecomment-4693968304, do we move `include/xsf/specfun.h` to `include/xsf/cpu/specfunh`?

steppi: I don't think that's necessary now, and is probably not going to reflect the eventual layout of xsf any way. #163 describes a longer term goal state, as we touch things, we should be mindful of whether or not we are moving further from or closer to the goal. There's no need to rush and try to restructure everything right now; we can just do it as we add more things to CuPy.

steppi:
> * Added `XSF_HOST_DEVICE` in cephes/round.h following @steppi 's suggestion

Sorry, I didn't mean to suggest for this to be done here. I just meant that there's no reason why that function shouldn't have `XSF_HOST_DEVICE`. It was an oversight not to include it.

steppi (approved):
This looks like it's ready to merge. Thanks @fbourgey!

PR #190, opened 2026-06-14
BUG(trig): sinpi/cospi: ensure all literals have correct type

### LINKED ISSUE #189: BUG(trig): sinpi/cospi: arithmetic not sensitive to input type
Hi!

I recently landed the addition of 5 "*pi" functions in CPython's `math` package. Now, I am interested in bringing cospi/sinpi to the Python API of numpy or scipy. There have been various discussions of where exactly this should land, but it led me to looking at the existing C++ implementation of `sinpi` and `cospi` in scipy/xsf.

I believe it does not correctly account for the varying argmuent types:
```c++
    template <typename T>
    XSF_HOST_DEVICE T cospi(T x) {
    // snipped: handle various reduction cases
        if (r < 1.0) {
            return -std::sin(M_PI * (r - 0.5));
```

Note how arithmetic is performed on the argument with literals like `0.5` and `M_PI`. To my understanding of C++, these literals always have type `double` and the result of `float * double` is `double`, leading to a call to `std::sin(double)` and ultimately a conversion of the return value back from a double to a float.

I believe a corrected implementation would look somewhat like this:
```c++
template <typename T>
T cospi(T x) {
    if (x < T(0.0)) {
        x = -x;
    }

    T r = std::fmod(x, T(2.0));
    if (r == T(0.5)) {
        // We don't want to return -0.0
        return T(0.0);
    }
    if (r < T(1.0)) {
        return -std::sin(T(M_PI) * (r - T(0.5)));
    } else {
        return std::sin(T(M_PI) * (r - T(1.5)));
    }
}
```

(though this is slightly incorrect for cospi(long double), because it does not use a long double version of the `pi` constant)

Before I go making a PR to fix this I wanted to check with the project and make sure my analysis is correct.

[godbolt test case](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGe1wAyeAyYAHI%2BAEaYxCAAzACcpAAOqAqETgwe3r56KWmOAkEh4SxRMQm2mPYFDEIETMQEWT5%2BXJXVGXUNBEVhkdFxiQr1jc05bcPdvSVlgwCUtqhexMjsHOaxwcjeWADUJrFuyCxMBAgH2CYaAIJX1wSYLEkGD/uHBACeSYysmLsAKhc7v9dmgFEk8BAQao5vsAOxWG67ZG7PBUXYQVRvNy7DQAOg0sJMCLuKLJuyxBwAIrsALSqA6I65k4lUu6klEg4hvGnDdAgEBUFgYTGkXZmAlzRkc5FojHc6nU3F4gCsRJJSPJyIA9NrdgB1P7oARgDgEXYAd0M5qIu2ImAIywYdPxGhl5PtjuIztd0s1yNZ7rlEAVh12XEl8KZWuRnqddL5ArSDAgAFkAPrKACSuwAVPK6cq1VLYtGUazdlUlFH3WS497domQMm05mc/mQ4WI8W/czy3C2TdA0Obg8ni8/gc3J9vsw2ACgTcQWCIVwoRT1WXZejMdiARBXXNN7WUZTYjT6b2WQP2f6AXaeY2CPzBcL0KL9xLCSWt6id6G2XPfd8WLGs7zJXUDSNE0zUta1dltetnVpX1wJRJDlTdUt3WHPttwLKd9wjb8wLwj0HXjWkmxbf5W2zWEO25WlgNVI8fxwgdK1oatiV/OsKIbajgihOiswYgtmNo7s2KvftB1uG8R2uKhaFQU5dioahVPUhpgE3O0BOdFc8DXXSf3heSVLU81gC06zdjMqMDK9IzUghCAzMZCyOAWWhOBVXg/A4LRSFQTg3GsaxGyWFZJzMWIeFIAhNB8hYEEwJgsBiCAFgAaziWJVTMAA2FUNDhAAOWJJDMeJ4jMMx9E4SRApS0LOF4BQQA0JKUoWOBYCQNAnjoaJyEoYaIXoGJgC4Yq2hoWgHmILqIAiNqImCBoPk4RLNuYYgPgAeQibRMAcNrhrYQQjoYWgduC3gsAiLxgDcMRuN2p7HkMYBxEe0h8HtBw8AAN0wLqAcwVRzq8B4vvIQQqja2g8AiYhto8LA2oIYg8BYBHweICJUkwKkfqMVGjD6vgDGABQADU8EwC0jtnBH%2BEEEQxHYKQZEERQVHUAHdEagxqdMSxrH0NGusgBZUCSGpIdpI6FF4VAibxrA5Zy9pzpqFwGHcTwWn8Y3pn6GI2jydIBDGVpklSO2GEt0oBgmKoDc6EYmlN8Z9ZBgQukaN3ZgmX2Hb0SZQ%2BCPp3ethYFBi1YJF8/zWoBsKOApCritpYrJF2YBkGQcNirxLgMVwQgSH2eKuDmXhkseo9SHSzKBj1/LYkKqQNC4FUKqq%2BLirhSRiqajgWtIAmuA0HqgpC7POu63rW9IAbEBAIhPHGjziHpjrSF0hRlEMKohAQVALSCxLJtG4hQl%2BThVDzgui5Lsu5sr3hMHwIg2s9Cc2EKIcQfMQGCzUG1UWs8BBMDQDXQBAAJJgCgLCYEYNcOGqA2aMH/gjYUzBEEAJIHgu%2BvAiEINQHUfAFDbDPmCMfPkwRz4hFoFfG%2B9Dd60CzPyBgLBSAWgxkkL66cOABVIEvDWnABEYlxl4BguVYRv3zoXYupdy6V2rqQhUDcxQeBGtNeusQzBN3XloNuPdCq91sXY%2BxU8Z7SPahwVePUW6WPEWYShEgF5SLaivCxqVSBEzSM4SQQA%3D)

jepler:
#### Reference issue
Closes: gh-189

 * Introduce and use `cephes/numbers.h`, an implementation of C++20 `<numbers>`, to get floating point constant values
 * In sinpi and cospi, always use values like `T(1.0)` or `pi_v<T>` so that intermediate calculations occur in the correct type

#### Additional information
Now, `sinpi(long double)` will use a long double precision approximation of pi. This may slightly change result values. Values for `sinpi(float)` might change too, because of no longer double rounding. However, when run locally, the tests passed on Linux.

The only functions I inspected were sinpi/cospi. It is likely that this class of error exists in other parts of the code. As suggested in #189, it might be good to try gcc `-Wfloat-conversion` to find them.

This is my first scipy/xsf contribution so please let me know what I can do to help move it along. Thanks!
 
#### AI Generation Disclosure
No AI tools used


include/xsf/cephes/numbers.h
```diff
@@ -0,0 +1,80 @@
+/*
+ * Mathematical constants
+ *
+ * This is the equivalent of the <numbers> header in C++20, but in the xsf::cephes namespace
+ */
+#pragma once
+
+#include <type_traits>
```
steppi: FYI, all `stdlib` headers need to be included through `xsf/config.h` to enable single-source use across CPU and GPU. Just remove `#include <type_traits>` and everything should work correctly. `std::enable_if` and `std::is_floating_point` will be available on both CPU and GPU after
`#include "../config.h"`.

include/xsf/cephes/trig.h
```diff
@@ -4,55 +4,126 @@
...
 
+    /* Compute acos(x) / pi */
+    template <typename T>
+    XSF_HOST_DEVICE T acospi(double x) {
+        double r = std::acos(x) / pi_v<T>;
+        if (std::isgreater(r, T(1.0))) {
+            return T(1.0);
+        }
+        return r;
+    }
```
steppi: Adding these new functions seems out of scope for this PR but would be welcome in a follow-up. 

Also, I don't think we want to use `std::isgreater` in `xsf`. `xsf` isn't intended as a standalone C++ library, but as a collection of numerical kernels for array-wise operations  in Python array libraries. There's an existing infrastructure in SciPy for [special function error handling](https://docs.scipy.org/doc/scipy/reference/special.html#error-handling) , floating point exceptions are ignored by default within NumPy universal functions (ufuncs), and users have the ability to control whether to ignore, warn, for different classes of exceptions. `xsf` offers hooks for other array libraries to set up their own error handling; currently only CuPy is also using `xsf`, and it unilaterally ignores all exceptions.

It would be safe and desirable to just use `return std::min(r, T(1.0);` below.

include/xsf/cephes/numbers.h
steppi, line 1: This is great! Setting up an equivalent of  `std::numbers` is something I've been wanting to do too

tests/xsf_tests/test_numbers.cpp
```diff
@@ -0,0 +1,12 @@
+#include "../testing_utils.h"
+#include <tuple>
+#include <xsf/cephes/numbers.h>
+
+TEST_CASE("numbers.h", "[numbers][xsf_tests]") {
+    REQUIRE(xsf::cephes::pi_v<float> == float(M_PI));
+    REQUIRE(xsf::cephes::pi_v<double> == double(M_PI));
+    REQUIRE(xsf::cephes::pi_v<long double> != double(M_PI));
```
steppi: Not all of the platforms `xsf` supports have an extended precision `long double`, as can be seen in the failing CI jobs.

steppi (changes requested):
Thanks @jepler! This is already looking very good. See the comments for the updates that will be needed. To summarize:

1. The new functions should be added in a separate PR
2. No direct standard library includes outside `config.h`, because we need a specific dance to enable single-source compilation on CPU and GPU.
3. Not all of the platforms we test on have an extended precision `long double`. I don't think think you need to test long doubles at all honestly. For the sake of portability, we don't really intend to support them.

include/xsf/cephes/numbers.h
j-bowhay: Should it be under `xsf::cephes` or just `xsf::detail`?

jepler:
Thanks for the comments. I'll work on addressing them during the coming week.

I'll for sure back out the additional functions in trig, that's out of scope for this PR and just included due to my inattention.

include/xsf/cephes/numbers.h
steppi: Good catch. It should just be in `xsf::detail`, we shouldn't be adding new things to the third-party sub-libraries.

include/xsf/cephes/trig.h
```diff
@@ -4,54 +4,55 @@
  */
 
 /*
- * Implement sin(pi * x) and cos(pi * x) for real x. Since the periods
+ * Implement half-turn trig functions for real x. Since the periods
  * of these functions are integral (and thus representable in double
- * precision), it's possible to compute them with greater accuracy
- * than sin(x) and cos(x).
+ * precision), it's possible to compute e.g., sinpi(x) them with
+ * greater accuracy than sin(pi * x).
```
jepler, line 10: this bit should also be dropped.

tests/xsf_tests/test_numbers.cpp
```diff
@@ -0,0 +1,16 @@
+#include "../testing_utils.h"
+#include <tuple>
+#include <xsf/numbers.h>
+
+TEST_CASE("numbers.h", "[numbers][xsf_tests]") {
+    REQUIRE(xsf::numbers::pi_v<float> == float(M_PI));
+    REQUIRE(xsf::numbers::pi_v<double> == double(M_PI));
+
+    // does not compile on all platforms, , is_floating_point<_Float16>::value is false
```
jepler, line 9: comment is wrong, revise it

jepler:
Do y'all prefer to rebase PRs as they go, or avoid rebasing them?

lucascolley:
> Do y'all prefer to rebase PRs as they go, or avoid rebasing them?

Hey @jepler, sorry for the long response, a rebase would be great. Let me know if you hit any issues

jepler:
Any advice on how to run the GPU tests locally? Here all tests fail with
```
E   cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorInsufficientDriver: CUDA driver version is insufficient for CUDA runtime version
```

lucascolley:
are you using `pixi run test-cupy`? If so, could you show the output of `pixi info`?

jepler:
Yes I am running `pixi run test-cupy`. All tests I inspected, including `TestCuPy.test_beta[tables_paths0]` which failed during ci fail for me with a similar error.

I included the full output of `pixi info` as well as some other info from my system that may or may not be relevant.

```System
------------
       Pixi version: 0.70.2
        TLS backend: rustls
           Platform: linux-64
   Virtual packages: __unix=0=0
                   : __linux=6.12.94=0
                   : __glibc=2.41=0
                   : __cuda=12.2=0
                   : __archspec=1=zen2
          Cache dir: /home/jepler/.cache/rattler/cache
       Auth storage: /home/jepler/.rattler/credentials.json
   Config locations: No config files found

Global
------------
            Bin dir: /home/jepler/.pixi/bin
    Environment dir: /home/jepler/.pixi/envs
       Manifest dir: /home/jepler/.pixi/manifests/pixi-global.toml

Workspace
------------
               Name: xsf
            Version: 0.2.2
      Manifest file: /home/jepler/src/xsf/pixi.toml
       Last updated: 15-06-2026 12:44:31

Environments
------------
        Environment: default
           Features: build, tests, tests-debug, default
        Solve group: default
           Channels: https://prefix.dev/conda-forge
   Dependency count: 5
       Dependencies: cmake, make, cxx-compiler, catch2, libarrow-all
   Target platforms: osx-64, linux-64, win-64, osx-arm64, linux-aarch64
    Prefix location: /home/jepler/src/xsf/.pixi/envs/default
              Tasks: configure-tests-debug, tests-debug, build-tests-debug, configure-tests, tests, build-tests, clean-xsref, clone-xsref, build, build-only, configure

        Environment: tests-ci
           Features: build, tests, tests-ci, tests-debug, tests-debug-ci, coverage, default
        Solve group: default
           Channels: https://prefix.dev/conda-forge
   Dependency count: 7
       Dependencies: cmake, make, cxx-compiler, catch2, libarrow-all, ccache, lcov
   Target platforms: linux-64, win-64, osx-64, linux-aarch64, osx-arm64
    Prefix location: /home/jepler/src/xsf/.pixi/envs/tests-ci
              Tasks: open-coverage, configure-coverage, tests-debug-ci, build-tests-debug-ci, configure-tests-debug, tests-debug, build-tests-debug, tests-ci, coverage, build-tests-ci, configure-tests, tests, build-tests, clean-xsref, clone-xsref, build, build-only, configure

        Environment: lint
           Features: clang-format, default
        Solve group: default
           Channels: https://prefix.dev/conda-forge
   Dependency count: 2
       Dependencies: git, clang-format
   Target platforms: osx-64, linux-64, win-64, linux-aarch64, osx-arm64
    Prefix location: /home/jepler/src/xsf/.pixi/envs/lint
              Tasks: format, format-dry-error

        Environment: cupy-tests
           Features: cupy-tests, default
        Solve group: default
           Channels: https://prefix.dev/conda-forge
   Dependency count: 6
       Dependencies: python, pip, setuptools, cupy, pytest, pytest-forked
   Target platforms: linux-64
    Prefix location: /home/jepler/src/xsf/.pixi/envs/cupy-tests
              Tasks: clone-xsref-test-cupy, install-xsref-test-cupy, test-cupy

```

```
$ nvidia-smi --query

==============NVSMI LOG==============

Timestamp                                 : Sun Sep 20 12:37:00 2026
Driver Version                            : 535.261.03
CUDA Version                              : 12.2

Attached GPUs                             : 1
GPU 00000000:09:00.0
    Product Name                          : NVIDIA GeForce RTX 3060
[snipped]
```

```
$ apt policy nvidia-cuda-dev nvidia-driver
nvidia-cuda-dev:
  Installed: 11.8.89~11.8.0-5~deb12u1
  Candidate: 11.8.89~11.8.0-5~deb12u1
  Version table:
 *** 11.8.89~11.8.0-5~deb12u1 500
        500 http://ftp.us.debian.org/debian bookworm/non-free amd64 Packages
        100 /var/lib/dpkg/status
nvidia-driver:
  Installed: 535.261.03-1
  Candidate: 535.261.03-1
  Version table:
 *** 535.261.03-1 500
        500 http://ftp.us.debian.org/debian bookworm/non-free amd64 Packages
        100 /var/lib/dpkg/status
```

jepler:
```
src/xsf$ pixi run test-cupy 
✨ Pixi task (clone-xsref-test-cupy in cupy-tests): bash -c '
if [ -d xsref ]; then
  commit=$(git -C xsref rev-parse HEAD 2>/dev/null || true)
fi
if [ "$commit" != 4e9b2343813136e355f0262754a3929d98bc5bff ]; then
  rm -rf xsref
  git clone https://github.com/scipy/xsref.git && git -C xsref checkout 4e9b2343813136e355f0262754a3929d98bc5bff
fi
': (Ensure xsref is present locally for CuPy tests.)

✨ Pixi task (install-xsref-test-cupy in cupy-tests): pip install .: (Install the local xsref package needed by CuPy tests.)
Processing /home/jepler/src/xsf/xsref
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: mpmath==1.3.0 in /home/jepler/src/xsf/.pixi/envs/cupy-tests/lib/python3.12/site-packages (from xsref==0.0.0) (1.3.0)
Requirement already satisfied: numpy in /home/jepler/src/xsf/.pixi/envs/cupy-tests/lib/python3.12/site-packages (from xsref==0.0.0) (2.2.6)
Requirement already satisfied: packaging in /home/jepler/src/xsf/.pixi/envs/cupy-tests/lib/python3.12/site-packages (from xsref==0.0.0) (25.0)
Requirement already satisfied: polars in /home/jepler/src/xsf/.pixi/envs/cupy-tests/lib/python3.12/site-packages (from xsref==0.0.0) (1.41.2)
Requirement already satisfied: pyarrow in /home/jepler/src/xsf/.pixi/envs/cupy-tests/lib/python3.12/site-packages (from xsref==0.0.0) (24.0.0)
Requirement already satisfied: scipy==1.15.2 in /home/jepler/src/xsf/.pixi/envs/cupy-tests/lib/python3.12/site-packages (from xsref==0.0.0) (1.15.2)
Requirement already satisfied: polars-runtime-32==1.41.2 in /home/jepler/src/xsf/.pixi/envs/cupy-tests/lib/python3.12/site-packages (from polars->xsref==0.0.0) (1.41.2)
Building wheels for collected packages: xsref
  Building wheel for xsref (pyproject.toml) ... done
  Created wheel for xsref: filename=xsref-0.0.0-py3-none-any.whl size=32994 sha256=beb865f550468596e344fa26a3e3a280ab80bf1ae6860bed808f18e070762650
  Stored in directory: /tmp/pip-ephem-wheel-cache-1z76n67p/wheels/02/6e/e7/bad8e6b1588fd8d76a48a4059f7938f85d0be27a8f519591c8
Successfully built xsref
Installing collected packages: xsref
  Attempting uninstall: xsref
    Found existing installation: xsref 0.0.0
    Uninstalling xsref-0.0.0:
      Successfully uninstalled xsref-0.0.0
Successfully installed xsref-0.0.0

✨ Pixi task (test-cupy in cupy-tests): pytest --forked python_tests/test_cupy.py: (Run CuPy integration tests.)
=============================================== test session starts ================================================
platform linux -- Python 3.12.11, pytest-8.4.0, pluggy-1.6.0
rootdir: /home/jepler/src/xsf
plugins: forked-1.6.0
collected 30 items                                                                                                 

python_tests/test_cupy.py FFFFFFFFxFFFFFFFFFFFFFFFFFFFFF                                                     [100%]

===================================================== FAILURES =====================================================
________________________________________ TestCuPy.test_beta[tables_paths0] _________________________________________
self = <test_cupy.TestCuPy object at 0x7f46fc48faa0>
tables_paths = (PosixPath('/home/jepler/src/xsf/xsref/tables/scipy_special_tests/beta/In_d_d-d.parquet'), PosixPath('/home/jepler/src...a/Out_d_d-d.parquet'), PosixPath('/home/jepler/src/xsf/xsref/tables/scipy_special_tests/beta/Err_d_d-d_other.parquet'))

    @pytest.mark.parametrize(
        "tables_paths", get_tables_for_func("beta")
    )
    def test_beta(self, tables_paths):
        input_path, output_path, tol_path = tables_paths
        beta = cupy._core.create_ufunc(
            "cupyx_scipy_beta",
            ("ll->d", "LL->d", "ee->d", "ff->f", "dd->d"),
            "out0 = out0_type(xsf::beta(in0, in1));",
            preamble=get_preamble("xsf/beta.h"),
        )
    
>       a, b = get_cols_as_cupy(input_path)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^

python_tests/test_cupy.py:150: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
python_tests/test_cupy.py:100: in get_cols_as_cupy
    return _get_cols_helper(table_path, cupy)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
python_tests/test_cupy.py:93: in _get_cols_helper
    result = tuple(xp.asarray(col) for col in zip(*rows))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
python_tests/test_cupy.py:93: in <genexpr>
    result = tuple(xp.asarray(col) for col in zip(*rows))
                   ^^^^^^^^^^^^^^^
.pixi/envs/cupy-tests/lib/python3.12/site-packages/cupy/_creation/from_data.py:94: in asarray
    return _core.array(a, dtype, copy, order, blocking=blocking)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
cupy/_core/core.pyx:2684: in cupy._core.core.array
    ???
cupy/_core/core.pyx:2704: in cupy._core.core.array
    ???
cupy/_core/core.pyx:2896: in cupy._core.core._array_default
    ???
cupy/_core/core.pyx:2849: in cupy._core.core._try_skip_h2d_copy
    ???
cupy/cuda/device.pyx:40: in cupy.cuda.device.get_device_id
    ???
cupy_backends/cuda/api/runtime.pyx:194: in cupy_backends.cuda.api.runtime.getDevice
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   cupy_backends.cuda.api.runtime.CUDARuntimeError: cudaErrorInsufficientDriver: CUDA driver version is insufficient for CUDA runtime version

cupy_backends/cuda/api/runtime.pyx:146: CUDARuntimeError
```

lucascolley:
Yeah sorry, the initial setup of the repo pre-dates Pixi having cuda-driver-specific platforms. Let me implement that today.

lucascolley:
gh-281

include/xsf/numbers.h
```diff
@@ -14,5 +14,71 @@ namespace numbers {
     template <>
     std::complex<double> i_v<double> = std::literals::complex_literals::operator""i(1.0L);
 
+    template <typename T>
```
lucascolley, line 22: Could you add a comment pointing to the reference for these numbers?

include/xsf/trig.h
```diff
@@ -28,7 +29,7 @@ XSF_HOST_DEVICE T sinpi(T x) {
 template <typename T>
 XSF_HOST_DEVICE std::complex<T> sinpi(std::complex<T> z) {
     T x = z.real();
-    T piy = M_PI * z.imag();
+    T piy = numbers::pi_v<T> * z.imag();
```
lucascolley, line 32: Are there uses of `M_PI` in other parts of the codebase that could/should be replaced?

include/xsf/trig.h
```diff
@@ -106,8 +107,8 @@ XSF_HOST_DEVICE std::complex<T> cospi(std::complex<T> z) {
         return {coshfac, sinhfac};
     }
 
-    coshfac = 0.5 * cospix * exphpiy;
-    sinhfac = 0.5 * sinpix * exphpiy;
+    coshfac = T(0.5) * cospix * exphpiy;
+    sinhfac = T(0.5) * sinpix * exphpiy;
```
lucascolley, line 111: I don't think it needs to block this PR, but maybe one could write a static lint check that looks out for bare literal usage inside templated funcs

tests/xsf_tests/test_numbers.cpp
lucascolley: looks like this is still TODO?

lucascolley (commented):
just a few comments, looks like you should run the formatter too

include/xsf/trig.h
jepler: This is quite likely to be the case. (254 hits for M_PI across the code base, not to mention other M_ math constants)

include/xsf/trig.h
jepler: If there's a desire to work on these issues system-wide, gcc -Wfloat-conversion is the tool I'd try to use.

include/xsf/trig.h
lucascolley: cool, let's open an issue to discuss further when this is merged

include/xsf/numbers.h
```diff
@@ -14,5 +14,71 @@ namespace numbers {
     template <>
     std::complex<double> i_v<double> = std::literals::complex_literals::operator""i(1.0L);
 
+    // These constants match llvm's libcxx <numbers> header
```
lucascolley:
```suggestion
    // These constants match llvm's libcxx <numbers> header
    // https://github.com/llvm/llvm-project/blob/466df937833ef60bc9f549a01059869ca7143d32/libcxx/include/numbers
```

lucascolley: BTW I don't think these match exactly? Some seem to have one extra digit, others one less

jepler: I double checked and they match (except that I have a trailing `l` to force the constants to long double precision; I'd hate to call llvm incorrect on a hunch, but [godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAKwAbKSuADJ4DJgAcj4ARpjEAaQADqgKhE4MHt6%2BAcGp6Y4C4ZExLPGJ/raY9kUMQgRMxATZPn5c1bWZDU0EJdFxCXoKjc2tuR0jvf1lFRIAlLaoXsTI7BwEmCzJBpsmAMxuyAYKCgDUACoH2CYaAIJoDCOYqsnEl2fJeAD6AG5nBwAImd9gA6KRcfwATjMgX8%2B38AA4oQB2KH7Mz7RGSQKwyT7LEYtH%2BDRmA5We6bba7TAHI4nc5XfY3e6PZ6vd4XT4/X5mAH7YFgiHQ2HwpGo9GY7G4wL4wlmYmk2jks4A%2B5TRzIb5MU4JAgQL5/Om0ATAM7oZaxejXfmAoHcv5kw4mozmy3W5nzckcRa0Tj%2BXh%2BDhaUioThuazWM4KZarTAAzE8UgETQ%2BxYITBMLCJCCLADWIAJoKxkml8IRUI0UP0nEkgdToc4vAUIA0ydTizgsBgiBQqG2dAS5EoaAH9ESwC4uL4dE2xBbEFiDdiESaAE9OEmV8xiGuAPKxbSYBwN0dsQR7hi0DfB3hYWJeYBuMS0FvcO9bQzAcS30j4YjHo4vyYG%2BIYvMeXibJuvARJsfq/rQeCxMQ64eFgDYEMQeAsNBpDAcQsRpJggKfkYiFGB2fAGMACgAGp4JgADue7JIwuH8IIIhiOwUgyIIigqOov66GY%2BhfigkaWPoSEtpAiyoMkdRvrwqD4VhWCybmnSAZkLgMO4nhtHoYQRAM5RDB0BQZAI4ztCkaTWQwMyDIkkw1DpAg9GMhkTNpDjdKMfSmbMFm2IFtnDIFznma5iwxisawLDWHABqQQYhmGHBnKoiKBAAtLKZzAMgyBnFOoJ8hAuCECQCb7Fw8y8Cmt7zOmmbZpQvq1rwOEkmlDaZc2rbti1pBdr2ywEMkkHDhAo5fOOUSsOsOX5YVxWleVSaYPgRDqXoHHCKI4i8YdAlqA2ImkIxKHJNBXUpfWv6ZXukHTQQZyoFQ2W5QVkhFSVZWBBVZwQB4Y4JHVZiNSNWitaQBZmJIoJQoiGiBIiZj%2BJIUKVvskhcPsyV1qQvVtulKlNrYw3NXD%2BaFmC%2BwaCi/gooEKIEmYGiSP4VTwfsT0ZVTtM%2BmNPbjRASDzYOZAUHN/YLUMxxfombY0LQc4Lkuv7buuuG67uB5Hie76kGejAEJe14Nvej7PrQr64VgLBfj%2BIb/jpwGgbw4HIJB6xJrBNQNohyGoRg6whph2G4fhhFKCRLtkREoCjVQ1F0QxzGsUGSaHVxJ3SGdSgXcJegGBRpiWNY0mxJp8mKZkymhmpeAafAcXuf5zgQK4EUdCZpQuXoVl1P39mFJk0VzG5XSeeFPl2XYHn1FFwXD5MC85Eva9DzFSXxXGSXwalFONllq1/Wcyuupi4KghooPVXtdUNU1HZtVmQxafBJNk/1z0qYtjbCLTs4skCTXerNaWi1lqcEvoVG%2BZo75cAfj7XaJA24HVkAXHiRdZDnSEiGXQHQbpMDuu%2BB6p8BqcFelNSCn1voIP%2Bkgs4KCH6g3Bord45h6ow1AfTIsTMWZsw5hibmvNiY9QCOTGhHAhogI/gjBmxZmas3ZpzCRfNOACwAULeRsM0zJTMNIrgGhZGAIMQIvCCR0jOEkEAA%3D%3D) thinks these two constants are not equal (tested gcc 16.2 and clang 23.0.1 targeting x86_64):
```
template<class T>
constexpr T pi_v = 3.141592653589793238462643383279502;
template<class T>
constexpr T pi_v2 = 3.141592653589793238462643383279502l;  
static_assert(pi_v<long double> == pi_v2<long double>);
```

gcc helpfully says
```
<source>:5:33: error: static assertion failed
    5 | static_assert(pi_v<long double> == pi_v2<long double>);
      |               ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
  • the comparison reduces to '(3.141592653589793116e+0l == 3.14159265358979323851e+0l)
```

jepler:
belatedly found this PR resolving the problem with clang libc++ pi_v<long double> et al: https://github.com/llvm/llvm-project/pull/222830

include/xsf/numbers.h
lucascolley: ah thanks, that must be what I was noticing

lucascolley (approved):
thanks @jepler, LGTM! We can open issues for following-up on other uses of `M_PI` and checking `gcc -Wfloat-conversion`.

jepler:
I appreciate the time you took with me on this PR!

jepler:
```
E           cupy.cuda.compiler.CompileException: /home/runnerx/actions-runner/_work/xsf/xsf/include/xsf/cephes/../numbers.h(12): error: name followed by "::" must be a class or namespace name
E                 std::complex<float> i_v<float> = std::literals::complex_literals::operator""if(1.0L);
```
that compiler error affecting lines I didn't modify has come back.

lucascolley:
> that compiler error affecting lines I didn't modify has come back.

yeah, I think `numbers.h` was not previously included in any files which CuPy uses, but it is now via `trig.h`. A fix may be to hide the definition behind `#ifndef __CUDACC__`

jepler:
I'm trying to reproduce this problem locally but I'm confused at what I'm seeing.

Is there an additional step I need to take between editing a header file and re-running `pixi run test-cupy`? In my local copy of `include/xsf/numbers.h` I wrote "#error I am a little teapot", which should make any inclusion of it into an error. But `pixi run test-cupy` still completes without any (unexpected) failed tests.

jepler:
also it's succeeded in CI at least once, https://github.com/scipy/xsf/actions/runs/35552327778/job/106281926537?pr=190 which doesn't look "substantially different" -- it has the inclusion of numbers.h from cepehes/trig.h which you suggested could be the cause of the build error.

lucascolley:
Apologies, see discussion at https://github.com/scipy/xsf/pull/280#issuecomment-5799240353. Try removing the kernel cache dir after edits and see if the behaviour is as expected?

We'll have to figure out the cache invalidation properly for CI, I think.

steppi:
> Apologies, see discussion at [#280 (comment)](https://github.com/scipy/xsf/pull/280#issuecomment-5799240353). Try removing the kernel cache dir after edits and see if the behaviour is as expected?
> 
> We'll have to figure out the cache invalidation properly for CI, I think.

Right. The only reason this failed this time but passed before is because the cache wasn't found this time

image

@lucascolley let's go back to not caching the kernels for now and do that ASAP. Then we can figure out how to handle caching properly at our leisure.

jepler:
Hey, this is also looking fishy:
```
    if (exphpiy == cxx::numeric_limits<T>::infinity()) {
```
isn't `==` always false when one argument is an infinity? shouldn't it be `!std::isfinite(exphpiy)` or something?

steppi:
> Hey, this is also looking fishy:
> 
> ```
>     if (exphpiy == cxx::numeric_limits<T>::infinity()) {
> ```
> 
> isn't `==` always false when one argument is an infinity? shouldn't it be `!std::isfinite(exphpiy)` or something?

I think you're thinking of `NaN`, not infinity. This seems fine. It's an explicit check for positive infinity. `!std::isfinite(exphpiy)` would be true for negative infinity and `NaN`, There's no way for `exphpiy` to be negative infinity here of course, but I don't think we want NaNs to go into this branch.

jepler:
> I think you're thinking of `NaN`, not infinity.

You're right. I was just sure that both behaved oddly with respect to equality but some experimentation on godbolt shows you're right.

PR #191, opened 2026-06-14
ENH: add cspence to xsf

j-bowhay:
#### Reference issue

towards  #170

adds https://github.com/scipy/scipy/blob/main/scipy/special/_spence.pxd
#### AI Generation Disclosure

I used copilot to review my code before submitting

### DIFF
--- a/include/xsf/spence.h
+++ b/include/xsf/spence.h
@@ -1,5 +1,6 @@
 /* Translated from Cython to C++ by SciPy developers in 2026.
  * Original authorship information below.
+ * 
  * Implement Spence's function, a.k.a. the dilogarithm, for complex
  * arguments. Note that our definition differs from that in the sources
  * by the mapping z -> 1 - z.
--- a/include/xsf/spence.h
+++ b/include/xsf/spence.h
@@ -13,6 +13,7 @@
 
 #pragma once
 
+#include "cephes/spence.h"
 #include "config.h"
 #include "zlog1.h"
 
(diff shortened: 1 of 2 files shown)


j-bowhay:
There is already an `xsf::cephes::spence`, I wasn't sure if I should add it as an overload there instead?

steppi:
Thanks @j-bowhay!

> There is already an `xsf::cephes::spence`, I wasn't sure if I should add it as an overload there instead?

The other way around actually. Add double and float overloads to `xsf/spence.h` that wrap `xsf::cephes::spence`.

j-bowhay:
> > There is already an `xsf::cephes::spence`, I wasn't sure if I should add it as an overload there instead?
> 
> The other way around actually. Add double and float overloads to `xsf/spence.h` that wrap `xsf::cephes::spence`.

That makes much more sense thanks

include/xsf/spence.h
```diff
@@ -0,0 +1,123 @@
+/*
```
steppi: We've been adding little notes at the top that say stuff like

```C++
/* Translated from Cython to C++ by SciPy developers in 2026.
 * Original authorship information below.

**Review by steppi — COMMENTED** [review]
Other than needing the little header comment about this being a translation from Cython, this looks good. Thanks @j-bowhay! At some point the tests for real valued `spence` functions should include and test the functions from the top level `xsf/spence.h`, and SciPy should use those too, but let's save it for later to minimize churn.

File: include/xsf/spence.h (hunk 1 above)
Comment (j-bowhay) [reply include/xsf/spence.h]:
```suggestion
/* Translated from Cython to C++ by SciPy developers in 2026.
 * Original authorship information below.
```

include/xsf/spence.h
```diff
@@ -0,0 +1,124 @@
+/* Translated from Cython to C++ by SciPy developers in 2026.
+ * Original authorship information below.
```
j-bowhay, line 2:
```suggestion
 * Original authorship information below.
 * 
```

PR #183, opened 2026-06-11
ENH: Add `hyperu` to XSF

fbourgey:
#### Reference issue
https://github.com/scipy/xsf/issues/170

Same pattern as https://github.com/scipy/xsf/pull/171

#### What does this implement/fix?
- C++ translation of [scipy/special/_hypergeometric.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_hypergeometric.pxd) to XSF

#### AI Generation Disclosure

I used Codex to draft a first translation version.

PR #192, opened 2026-06-14
MAINT: move `hyperu` to `xsf/cpu` and add float overload

fbourgey:
- `hyperu` was recently added to xsf in https://github.com/scipy/xsf/pull/183. It seems that it should have been placed into `xsf/cpu` as `hypu` does not have XSF_HOST_DEVICE. This PR does this.
- Add also missing float overload for hyperu.

#### AI Generation Disclosure

No AI.


steppi:
This isn't really necessary yet because there is no device-capable `xsf/hyperu.h` header we need to keep `xsf/cpu/hyperu.h`  separate from. I suspect `hyperu` should actually work on GPU too, the issue is that it depends on `xsf/specfun.h` and no one has separated out the things that will work on GPU from those which don't in `specfun`. Unless you are faced with a concrete situation where you want to include a header in a CuPy ufunc but there are host only things mixed into the header causing compilation errors, I wouldn't worry about splitting things out like this right now.

fbourgey:
> This isn't really necessary yet because there is no device-capable `xsf/hyperu.h` header we need to keep `xsf/cpu/hyperu.h` separate from. I suspect `hyperu` should actually work on GPU too, the issue is that it depends on `xsf/specfun.h` and no one has separated out the things that will work on GPU from those which don't in `specfun`. Unless you are faced with a concrete situation where you want to include a header in a CuPy ufunc but there are host only things mixed into the header causing compilation errors, I wouldn't worry about splitting things out like this right now.

I see, thanks Albert. Closing this one and i'll open another one for the float overload.

PR #193, opened 2026-06-14
ENH: add float overload for `hyperu`

fbourgey:
Add missing float overload for `hyperu`.

#### AI Generation Disclosure

No AI.

### DIFF
--- a/include/xsf/hyperu.h
+++ b/include/xsf/hyperu.h
@@ -36,4 +36,8 @@ inline double hyperu(double a, double b, double x) {
     return hypu(a, b, x);
 }
 
+inline float hyperu(float a, float b, float x) {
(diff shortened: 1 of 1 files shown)


include/xsf/spence.h
```diff
@@ -1,6 +1,6 @@
 /* Translated from Cython to C++ by SciPy developers in 2026.
  * Original authorship information below.
- * 
+ *
```
steppi, line 3: This isn't really in-scope for this PR. I'd suggest doing it separately, even if the format job fails without it. @j-bowhay, don't use `[skip-ci]` again for changes like this, we at least need to run the format check.

fbourgey: @steppi it should be good now. I self-merged https://github.com/scipy/xsf/pull/194 as this was minimal, and force-pushed in here.

steppi (approved):
Looks good. Thanks @fbourgey

PR #196, opened 2026-06-16
Remove constexpr uses of QuietNaN to enable nvrtc 120 support

steppi:

This PR removes a `constexpr` alias for `std::numeric_limits<T>::QuietNaN()` instead inserting this value directly where it is needed. The purpose of this PR is to unblock https://github.com/cupy/cupy/pull/10005. It was found in CI there that `constexpr` `QuietNaN` is not supported with `nvrtc 120` despite it having been part of the standard since C++11.


j-bowhay (approved):

steppi:
~Self merging due to relative triviality~ Thanks @j-bowhay!

PR #197, opened 2026-06-17
ENH: Add `hyp0f1` to XSF

fbourgey:
#### Reference issue

- C++ translation of [scipy/special/_hyp0f1.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/_hyp0f1.pxd) to XSF

#### AI Generation Disclosure


include/xsf/hyp0f1.h
```diff
@@ -0,0 +1,130 @@
...
+    if (std::abs(z) < 1e-6 * (1.0 + std::abs(v))) {
+        return 1.0 + z / v + z * z / (2.0 * v * (v + 1.0));
+    }
+
+    if (z > 0.0) {
+        arg = std::sqrt(z);
+        arg_exp = xsf::xlogy(1.0 - v, arg) + xsf::gammaln(v);
+        bess_val = xsf::cyl_bessel_i(v - 1.0, 2.0 * arg);
+
+        if (arg_exp > std::log(std::numeric_limits<double>::max()) || bess_val == 0.0 ||      // overflow
```
dschmitz89, line 79: Can we include log(max_double) and log(min_douboe) in const.h if they are not already?

dschmitz89 (commented):
One minor inline comment and one suggestion: let's add at least one test with float precision.

include/xsf/hyp0f1.h
fbourgey: We can use MAXLOG from `cephes/const.h` but I think `MINLOG` is not the same as log(min_double).

Is adding/renaming old cephes constants what we want to do here? I see there is also a `xsf/numbers.h` @steppi 

```python
import sys

import mpmath as mp

mp.mp.dps = 80

dbl_max = mp.mpf(sys.float_info.max)
dbl_min_normal = mp.mpf(sys.float_info.min)

print(mp.log(dbl_max))
print(mp.log(dbl_min_normal))

print("MAXLOG const:")
print(mp.mpf("7.097827128933839730962063185871E2"))

print("MINLOG const:")
print(mp.mpf("-7.451332191019412076235245305675398106812E2"))
```

gives

```python
709.78271289338399673222338991065714550397314873666416303860305771470605931916868
-708.396418532264106224411228130256452573161137316180869711334969702248281653028
MAXLOG const:
709.7827128933839730962063185871
MINLOG const:
-745.1332191019412076235245305675398106812
```

fbourgey:
> One minor inline comment and one suggestion: let's add at least one test with float precision.

@dschmitz89 thanks for the review. I added some float overload tests.

include/xsf/hyp0f1.h
steppi: Yeah, I think the preferred thing should be to add new constants to `xsf/numbers.h`

PR #198, opened 2026-06-18
ENH: Add CDF and SF of the cosine distribution to XSF

fbourgey:
#### Reference issue

- C++ translation of [scipy/special/_cosine.c](https://github.com/scipy/scipy/blob/main/scipy/special/_cosine.c) to XSF

#### AI Generation Disclosure

### DIFF
--- a/include/xsf/cosine.h
+++ b/include/xsf/cosine.h
@@ -1,0 +1,229 @@
+// Translated from Cython to C++ by the xsf developers in 2026.
+#pragma once
+
+#include "alg.h"
+#include "cephes/polevl.h"
+#include "config.h"
+
+namespace xsf {
+
+namespace detail {
+
+    // M_PI64 is the 64 bit floating point representation of π, e.g., in Python:
+    //   >>> import math
+    //   >>> math.pi.hex()
+    //   '0x1.921fb54442d18p+1'
+    // It is used in the function cosine_cdf_pade_approx_at_neg_pi,
+    // which depends on this value being the 64 bit representation.
+    // Do not replace this with M_PI from math.h or NPY_PI from the
+    // numpy header files.
+    inline constexpr double M_PI64 = 0x1.921fb54442d18p+1;
+
+    // p and q (below) are the coefficients in the numerator and denominator
+    // polynomials (resp.) of the Pade approximation of
+    //     f(x) = (pi + x + sin(x))/(2*pi)
+    // at x=-pi.  The coefficients are ordered from lowest degree to highest.
+    // These values are used in the function cosine_cdf_pade_approx_at_neg_pi(x).
+    //
+    // These coefficients can be derived by using mpmath as follows:
+    //
+    //    import mpmath
+    //
+    //    def f(x):
+    //        return (mpmath.pi + x + mpmath.sin(x)) / (2*mpmath.pi)
+    //
+    //    # Note: 40 digits might be overkill; a few more digits than the default
+    //    # might be sufficient.
+    //    mpmath.mp.dps = 40
+    //    ts = mpmath.taylor(f, -mpmath.pi, 20)
+    //    p, q = mpmath.pade(ts, 9, 10)
+    //
+    // (A python script with that code is in special/_precompute/cosine_cdf.py.)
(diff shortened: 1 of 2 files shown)


include/xsf/cosine.h
```diff
@@ -0,0 +1,229 @@
...
+    //    def f(x):
+    //        return (mpmath.pi + x + mpmath.sin(x)) / (2*mpmath.pi)
+    //
+    //    # Note: 40 digits might be overkill; a few more digits than the default
+    //    # might be sufficient.
+    //    mpmath.mp.dps = 40
+    //    ts = mpmath.taylor(f, -mpmath.pi, 20)
+    //    p, q = mpmath.pade(ts, 9, 10)
+    //
+    // (A python script with that code is in special/_precompute/cosine_cdf.py.)
```
dschmitz89, line 41: Can we include this here as well?

steppi: Scripts for generating reference values belong in `xsref`

WarrenWeckesser: My impression was that xsref is for test cases. This little script is for generating coefficients that are used in the code.  I don't think xsf has a place yet for storing code-generation code.

steppi: Beyond test cases it also has test case generation scripts. I also wanted to add things like workflows to get minimax polynomials with remez exchange to what is now `xsref`. Beyond just test cases and test case generation I think it could be a home for the "meta-code" that goes into `xsf`. `xsref` is due for a namechange though and probably `xsf` too since I hope to expand the scope beyond special functions

fbourgey: > My impression was that xsref is for test cases. This little script is for generating coefficients that are used in the code. I don't think xsf has a place yet for storing code-generation code.

This is what I had in mind too and this is why I kept it from the original Cython file. @steppi what are you suggesting here? should we wait before moving those comments over to xsref or we do it now?

steppi: Sorry for the noise. I thought you were all talking about adding the Python script to the `xsf` repo, not the comment. I'm on leave and don't have much availability now and trust you all to figure out what's best to do here

WarrenWeckesser: Heh, well, *I* was thinking about adding the Python script that generates the coefficients to the `xsf` repo.  That's where it makes the most sense to me (as opposed to `xsref`, which I have been seeing as a repo related to test data, including code that generates reference values for unit tests) . But as long as the code is *somewhere*--even in a comment--then it can be used again by future devs, and they don't have to wonder how the rational approximation was derived.

dschmitz89: I agree with @WarrenWeckesser here. The script that contains the generation of the coefficients belongs in the repo that contains the implementation. Can we agree on adding it here for the time being and potentially move it to `xsfref` once that is more mature? Otherwise, this PR is mergeable.

fbourgey: Sounds good to me.

dschmitz89:
@fbourgey gentle ping

fbourgey:
> @fbourgey gentle ping

@dschmitz89 This is fine with me. I let you merge if OK with you as I opened the PR.

dschmitz89:
@fbourgey I thought we converged on adding the script for the derivation of the coefficients here still: see https://github.com/scipy/xsf/pull/198#discussion_r3466956277 to avoid losing it.

fbourgey:
I think I am missing something.. The script to generate the Padé coefficients is already in the comments in this PR (from line 28 to 40) Are you thinking of adding something else?

https://github.com/scipy/scipy/blob/main/scipy/special/_precompute/cosine_cdf.py

dschmitz89:
> I think I am missing something.. The script to generate the Padé coefficients is already in the comments in this PR (from line 28 to 40) Are you thinking of adding something else?
> 
> https://github.com/scipy/scipy/blob/main/scipy/special/_precompute/cosine_cdf.py

You are absolutely right, it was me who missed something, sorry. Thanks for the PR, merging now.

dschmitz89 (approved):

PR #199, opened 2026-06-19
ENH: migrate wright to xsf

j-bowhay:
#### Reference issue

Towards #170

Migrates `wright.c` to xsf.
#### AI Generation Disclosure

I used copilot to convert the tests from Python.

### DIFF
--- a/include/xsf/wright.h
+++ b/include/xsf/wright.h
@@ -392,7 +392,7 @@ XSF_HOST_DEVICE inline double wrightomega(double x) {
 
     if (x < -50.0) {
         /*
-         * Skip the iterative scheme because it exp(x) is already
+         * Skip the iterative scheme because  exp(x) is already
          * accurate to double precision.
          */
         w = std::exp(x);
@@ -409,7 +409,7 @@ XSF_HOST_DEVICE inline double wrightomega(double x) {
         return x;
     }
 
-    /* Split int three distinct intervals (-inf,-2), [-2,1), [1,inf) */
+    /* Split into three distinct intervals (-inf,-2), [-2,1), [1,inf) */
     if (x < -2.0) {
         /* exponential is approx < 1.3e-1 accurate */
         w = std::exp(x);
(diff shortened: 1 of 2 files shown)


include/xsf/wright.h
```diff
@@ -0,0 +1,444 @@
...
+        /***********************/
+        /* Undo regularization */
+        /***********************/
+        *w = s * *w;
+
+        /***************************************************/
+        /*  Provide condition number estimate if requested */
+        /***************************************************/
+        if (cond != NULL) {
+            *cond = z / (1.0 + *w);
```
dschmitz89, line 356: Is this condition number ever used anywhere?

include/xsf/wright.h
```diff
@@ -0,0 +1,444 @@
...
+        if (x > 0.0) {
+            return x;
+        } else {
+            return 0.0;
+        }
+    }
+
+    if (x < -50.0) {
+        /*
+         * Skip the iterative scheme because it exp(x) is already
```
dschmitz89:
```suggestion
         * Skip the iterative scheme because  exp(x) is already
```

include/xsf/wright.h
```diff
@@ -0,0 +1,444 @@
...
+    }
+    if (x > 1e20) {
+        /*
+         * Skip the iterative scheme because the result is just x to
+         * double precision
+         */
+        return x;
+    }
+
+    /* Split int three distinct intervals (-inf,-2), [-2,1), [1,inf) */
```
dschmitz89:
```suggestion
    /* Split into three distinct intervals (-inf,-2), [-2,1), [1,inf) */
```

include/xsf/wright.h
j-bowhay: No it isn't. I wasn't sure what the procedure for migrating stuff is, i.e. I have made as few modifications as possible

dschmitz89: Alright, let's leave it for a follow up then. First priority is to get this function into xsf in its current shape.

Issue #200, opened 2026-06-19
CI: check ccache configuration


lucascolley:
I suspect there is some misconfiguration of ccache — compare for example the following two jobs, for which the 'Build xsf' step is very quick:

- https://github.com/lucascolley/xsf/actions/runs/27825553418/job/82348925291?pr=41
- https://github.com/lucascolley/xsf/actions/runs/27825553418/job/82348925283?pr=41

Against these two, for which the same step takes upwards of five minutes:

- https://github.com/scipy/xsf/actions/runs/27826566797/job/82352380830?pr=41
- https://github.com/scipy/xsf/actions/runs/27826566797/job/82352380892?pr=41

@nickodell is it problematic that the keys at https://github.com/scipy/xsf/blob/6170343b2f1c335c20be578cfa8bde55661a241f/.github/workflows/tests.yml#L45-L46 do not differentiate between the 'Release' and 'Debug' jobs? That might explain why the ubuntu debug job takes much longer, but does not seem to explain why the windows job takes so long?

There are other things done in the `scipy` repo that aren't done here, e.g. https://github.com/scipy/scipy/blob/fd362f891a909ea3f80ea67fc5c180a53afa23d1/.github/workflows/macos.yml#L18 .

nickodell:
@lucascolley

- Yeah, that looks problematic to me. A file compiled with different flags can't be re-used. Would recommend putting the job name into the key in addition to the workflow name that's already there.
  
  (I have tooling in actions-search to look for cross-job cache restores, but it's not running on `scipy/xsf` job logs. I could do that, if it would be helpful.)

- It's not the only problem at play though. A second issue is that in [this job](https://github.com/scipy/xsf/actions/runs/27826566797/job/82352380830?pr=41#step:11:1). you can see that according to `ccache -s`, no use of ccache was even attempted. meson usually autoconfigures `ccache`, but I don't know if you use meson in this project? I also don't know if cmake autoconfigures `ccache`.
 
- Thirdly, any pixi job needs [`CCACHE_COMPILERCHECK=hash`](https://github.com/scipy/scipy/blob/fd362f891a909ea3f80ea67fc5c180a53afa23d1/.github/workflows/linux.yml#L20) for the cache to re-used between jobs. 

   It's apparently getting 50% hits without it; either the build is being run twice, or pixi restoring its dependencies from cache with tar preserves mtimes.

lucascolley:
[...]

Issue #31, opened 2025-04-20
BLD: MSVC portability


lucascolley:
@steppi are you still having to add extra args to the build invocation to make things work with VS2022?

I could reproduce the same error you had previously, but not anymore. I added some extra components to my VS installation, which might have made the difference. Here's my full list of components:

![Image](https://github.com/user-attachments/assets/1c10a46f-dab0-4bf6-9bc2-7f8e23458727)

![Image](https://github.com/user-attachments/assets/fd4aa4ed-2fa0-4c9d-91cd-6f253f39e6fb)

If I had to guess, `MSVC v142 - VS 2019 C++ x64/x86 build tool...` might be it.

---

Either way, it seems desirable to make things maximally portable across VS installations. It was suggested at https://discord.com/channels/1082332781146800168/1360605502169681940/1363258900261900350 to apply a diff like 

```diff
diff --git a/pixi.toml b/pixi.toml
index 2be7f18..09bb43d 100644
--- a/pixi.toml
+++ b/pixi.toml
@@ -16,6 +16,10 @@ cmake = ">=3.30.5,<4"
 cxx-compiler = ">=1.8.0,<2"
 make = ">=4.4.1,<5"

+[feature.build.target.win-64.dependencies]
+vs2019_win-64 = "*"
+vs2022_win-64 = "*"
+
 [feature.build.tasks.configure]
 cmd = [
   "cmake",
```

and perhaps use `ninja`?

(It was also explained there that `cxx-compiler` is an alias for `vs2019_win-64` on windows, which explains why CMake was expecting VS2019 in some sense. Thus the addition of `vs2022_win-64` would perhaps be sufficient.)

lucascolley:
may also be interesting to see if `clang-cl` works

PR #203, opened 2026-06-21
CI: ccache fixes

lucascolley:
towards gh-200. Not figured out the windows problem yet.


pixi.toml
```diff
@@ -18,11 +18,7 @@ preview = ["pixi-build"]
 # environments
 
 [environments.default]
-features = ["build", "tests", "tests-debug"]
-solve-group = "default"
-
-[environments.ci]
-features = ["build", "tests", "tests-debug", "cov", "ci"]
+features = ["build", "tests", "tests-debug", "cov"]
 solve-group = "default"
```
lucascolley, line 22: let's just simplify this and have one env

lucascolley:
@fbourgey I'm inclined to merge this to speed up unix, then investigate using `ninja` (previously mentioned in gh-31) before debugging getting ccache working on windows. The linked issue can be renamed to be specific to windows. WDYT?

lucascolley:
@nickodell how are these ccache stats looking?

nickodell:
> @nickodell how are these ccache stats looking?

1. [MacOS hit rate seems low:](https://github.com/scipy/xsf/actions/runs/27906604813/job/82576275996#step:11:10)

   ```
     Hits:             203 /  609 (33.33%)
   ```

   Typically, ccache can hit 80%-100%. Here's a histogram of scipy/scipy's typical ccache hit rate, across all jobs. The pattern here is that you can see most jobs have >95% hit rate. Some jobs fail to restore a cache entry or are misconfigured, so they get a 0% hit rate.

   newplot

2. To have a valid test, you really need to both generate and consume the cache entry under the new configuration. I would suggest changing the prefix of all of the cache keys to prevent it from getting a cache hit from main, then running these tests through CI twice.

   The MacOS hit rate thing I mentioned could just be that CCACHE_COMPILERCHECK changed.

lucascolley:
After changing the key prefix, the jobs are still at exactly 203 hits (50%). Sounds like there is something else going on @nickodell ?

lucascolley:
going to merge this to speed up unix CI for everyone substantially, will continue looking into windows in a follow-up

PR #204, opened 2026-06-22
ENH: add NaN guardrails for `hyp0f1`

fbourgey:
Add NaN guardrails for hyp0f1 after https://github.com/scipy/scipy/pull/25467 is failing

#### AI Generation Disclosure

No AI

### DIFF
--- a/include/xsf/hyp0f1.h
+++ b/include/xsf/hyp0f1.h
@@ -58,6 +58,10 @@ namespace detail {
 inline double hyp0f1(double v, double z) {
     double arg, arg_exp, bess_val;
 
+    if (std::isnan(v) || std::isnan(z)) {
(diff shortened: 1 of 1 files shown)


dschmitz89 (approved):

fbourgey:
Thanks Daniel.

dschmitz89:
@fbourgey : I think it makes sense to have a generic solution for this that works for all functions to avoid repetition. What do you think about a small helper function that simply returns Nan if any input Parameter is Nan?

fbourgey:
> @fbourgey : I think it makes sense to have a generic solution for this that works for all functions to avoid repetition. What do you think about a small helper function that simply returns Nan if any input Parameter is Nan?

Sounds like a good idea.

PR #208, opened 2026-06-29
BLD: use `ninja` instead of `make`

lucascolley:
related to gh-31

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -68,21 +68,15 @@ cmd = [
 
 [feature.build.tasks.build-only]
 description = "Build the release CMake configuration in the build directory."
-args = [{ arg = "parallel", default = "2" }]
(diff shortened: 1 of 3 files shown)


pixi.toml
```diff
@@ -97,28 +93,21 @@ cmd = [
...
-  "configure-tests",
-  { task = "build-only", args = ["{{ parallel }}"] },
-]
+depends-on = ["configure-tests", "build-only"]
 
 [feature.tests.tasks.tests-only]
 description = "Run CTest test suite from build/tests with verbose failures."
-args = [{ arg = "parallel", default = "2" }]
-cmd = ["ctest", "--output-on-failure", "--test-dir", "build/tests", "-j{{ parallel }}"]
+cmd = ["ctest", "--output-on-failure", "--test-dir", "build/tests"]
```
lucascolley: oops we probably shouldn't get rid of this for `ctest` even if we decide to get rid of it for the build

PR #209, opened 2026-07-01
ENH: Add `jacobi` and `shifted jacobi` to XSF

khushi-411:
#### Reference issue

Follows: #170

- C++ port of Jacobi and Shifted Jacobi in [scipy/special/orthogonal_eval.pxd](https://github.com/scipy/scipy/blob/main/scipy/special/orthogonal_eval.pxd) to XSF

#### AI Generation Disclosure

I used Augment to draft the test cases. The rest is mine.

### DIFF
--- a/include/xsf/orthogonal_eval.h
+++ b/include/xsf/orthogonal_eval.h
@@ -1,0 +1,94 @@
+#pragma once
+
+#include "binom.h"
+#include "config.h"
+#include "hyp2f1.h"
+
+namespace xsf {
+
+// Jacobi
+
+template <typename T>
+XSF_HOST_DEVICE inline T eval_jacobi(double n, double alpha, double beta, T x) {
+    double a, b, c, d;
+    T g;
+
+    if (alpha == -1 and std::abs(beta) == 1) {
+        if (n == 0) {
+            return 1.0;
+        } else if (n == 1) {
+            return 0.5 * (1.0 + beta) * (x - 1.0);
+        } else if (n > 1) {
+            return ((n + beta) / (2.0 * n)) * (x - 1) * eval_jacobi(n - 1, 1, beta, x);
+        }
+    }
+
+    d = binom(n + alpha, n);
+    a = -n;
+    b = n + alpha + beta + 1;
+    c = alpha + 1;
+    g = 0.5 * (1 - x);
+    return d * hyp2f1(a, b, c, g);
+}
+
+XSF_HOST_DEVICE inline double eval_jacobi_l(int n, double alpha, double beta, double x) {
+    int kk;
+    double p, d;
+    double k, t;
+
+    if (n < 0) {
+        return eval_jacobi(n, alpha, beta, x);
+    } else if (n == 0) {
+        return 1.0;
+    } else if (n == 1) {
+        return 0.5 * (2 * (alpha + 1) + (alpha + beta + 2) * (x - 1));
+    } else if (alpha == -1 and std::abs(beta) == 1) {
+        return ((n + beta) / (2.0 * n)) * (x - 1) * eval_jacobi(n - 1, 1, beta, x);
+    } else {
+        d = (alpha + beta + 2) * (x - 1) / (2 * (alpha + 1));
+        p = d + 1;
+        for (kk = 0; kk < n - 1; kk++) {
+            k = kk + 1.0;
+            t = 2 * k + alpha + beta;
+            d = ((t * (t + 1) * (t + 2)) * (x - 1) * p + 2 * k * (k + beta) * (t + 2) * d) /
+                (2 * (k + alpha + 1) * (k + alpha + beta + 1) * t);
+            p = d + p;
+        }
+        return binom(n + alpha, n) * p;
+    }
+}
+
+template <typename T>
+XSF_HOST_DEVICE inline float eval_jacobi(float n, float alpha, float beta, T x) {
+    return eval_jacobi(
+        static_cast<double>(n), static_cast<double>(alpha), static_cast<double>(beta), static_cast<double>(x)
+    );
+}
+
+XSF_HOST_DEVICE inline float eval_jacobi_l(int n, float alpha, float beta, float x) {
+    return eval_jacobi_l(n, static_cast<double>(alpha), static_cast<double>(beta), static_cast<double>(x));
+}
+
+// Shifted Jacobi
+
+template <typename T>
+XSF_HOST_DEVICE inline T eval_sh_jacobi(double n, double p, double q, T x) {
+    return eval_jacobi(n, p - q, q - 1, 2 * x - 1) / binom(2 * n + p - 1, n);
+}
+
+XSF_HOST_DEVICE inline double eval_sh_jacobi_l(int n, double p, double q, double x) {
+    return eval_jacobi_l(n, p - q, q - 1, 2 * x - 1) / binom(2 * n + p - 1, n);
+}
+
+template <typename T>
+XSF_HOST_DEVICE inline T eval_sh_jacobi(float n, float p, float q, T x) {
+    return eval_sh_jacobi(
+        static_cast<double>(n), static_cast<double>(p), static_cast<double>(q), static_cast<double>(x)
+    );
+}
+
+XSF_HOST_DEVICE inline float eval_sh_jacobi_l(int n, float p, float q, float x) {
+    return eval_sh_jacobi_l(n, static_cast<double>(p), static_cast<double>(q), static_cast<double>(x));
+}
+
+} // namespace xsf
--- a/include/xsf/orthogonal_eval.h
+++ b/include/xsf/orthogonal_eval.h
@@ -31,8 +31,8 @@ namespace detail {
         return d * hyp2f1(a, b, c, g);
     }
 
-    XSF_HOST_DEVICE inline double eval_jacobi_l(long n, double alpha, double beta, double x) {
-        long kk;
+    XSF_HOST_DEVICE inline double eval_jacobi_l(std::ptrdiff_t n, double alpha, double beta, double x) {
+        std::ptrdiff_t kk;
         double p, d;
         double k, t;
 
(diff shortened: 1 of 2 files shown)


dschmitz89:
Thanks for the PR. I only took a glance so far. Before we run tests in CI, could you reimplement the corresponding tests from SciPy for these functions? That's what we did so far with these migrations to be on the safe side

khushi-411:
Hi @dschmitz89, thanks for taking a look at the PR! I've now reimplemented the corresponding tests based on [scipy/special/tests/test_orthogonal_eval.py](https://github.com/scipy/scipy/blob/main/scipy/special/tests/test_orthogonal_eval.py). Please let me know if there's anything else you'd like me to update.

include/xsf/orthogonal_eval.h
```diff
@@ -0,0 +1,94 @@
...
+#include "binom.h"
+#include "config.h"
+#include "hyp2f1.h"
+
+namespace xsf {
+
+// Jacobi
+
+template <typename T>
+XSF_HOST_DEVICE inline T eval_jacobi(double n, double alpha, double beta, T x) {
```
dschmitz89: I cross checked this with the SciPy source and get confused. That might also be because of the whole `scipy.special` wrapper xomplexity.

Acc. to what we currently have in SciPy, this function should support double and double complex: see [here](https://github.com/scipy/scipy/blob/8e93e0478ca5b6e0b51652a1395f54160be2a672/scipy/special/orthogonal_eval.pxd#L72) and [here](https://github.com/scipy/scipy/blob/8e93e0478ca5b6e0b51652a1395f54160be2a672/scipy/special/cython_special.pxd#L2). I do not see tests with double complex input though. Are there none in SciPy either? 

I also cannot see an explicity complex overload as a public function here. SciPy contains one though: https://github.com/scipy/scipy/blob/8e93e0478ca5b6e0b51652a1395f54160be2a672/scipy/special/functions.json#L253

khushi-411: > Acc. to what we currently have in SciPy, this function should support double and double complex: see [here](https://github.com/scipy/scipy/blob/8e93e0478ca5b6e0b51652a1395f54160be2a672/scipy/special/orthogonal_eval.pxd#L72) and [here](https://github.com/scipy/scipy/blob/8e93e0478ca5b6e0b51652a1395f54160be2a672/scipy/special/cython_special.pxd#L2).

This was indeed informative. I wasn't aware that `number_t` was specifically limited to `double` and `std::complex<double>`. Thanks for the links, @dschmitz89! I've added overloads and tests for `std::complex<double>` inputs.

> I do not see tests with double complex input though. Are there none in SciPy either?

Yes, I didn't find any tests for complex inputs in SciPy, I looked through `scipy/special/tests/test_orthogonal_eval.py`, [here](https://github.com/scipy/scipy/blob/f4dfd5496b357fee36967e23920d833d204131ad/scipy/special/tests/test_orthogonal_eval.py#L77-L85), [here](https://github.com/scipy/scipy/blob/f4dfd5496b357fee36967e23920d833d204131ad/scipy/special/tests/test_orthogonal_eval.py#L185-L192), [here](https://github.com/scipy/scipy/blob/f4dfd5496b357fee36967e23920d833d204131ad/scipy/special/tests/test_orthogonal_eval.py#L322) and [here](https://github.com/scipy/scipy/blob/f4dfd5496b357fee36967e23920d833d204131ad/scipy/special/tests/test_orthogonal_eval.py#L376), but couldn't find any. 

Would you like me to add complex tests there? If so, I'd be happy to add them in my next PR.

> I also cannot see an explicity complex overload as a public function here. SciPy contains one though: https://github.com/scipy/scipy/blob/8e93e0478ca5b6e0b51652a1395f54160be2a672/scipy/special/functions.json#L253

Updated, thanks!

include/xsf/orthogonal_eval.h
```diff
@@ -0,0 +1,111 @@
+#pragma once
+
+#include <cmath>
+#include <complex>
+#include <cstddef>
```
j-bowhay:
```suggestion
```

These are already included via `config.h`

include/xsf/orthogonal_eval.h
```diff
@@ -0,0 +1,111 @@
...
+} // namespace detail
+
+// Jacobi
+
+XSF_HOST_DEVICE inline double eval_jacobi(double n, double alpha, double beta, double x) {
+    return detail::eval_jacobi_impl(n, alpha, beta, x);
+}
+
+XSF_HOST_DEVICE inline float eval_jacobi(float n, float alpha, float beta, float x) {
+    return eval_jacobi(
```
j-bowhay:
```suggestion
    return detail::eval_jacobi(
```

include/xsf/orthogonal_eval.h
```diff
@@ -0,0 +1,111 @@
...
+        static_cast<double>(n), static_cast<double>(alpha), static_cast<double>(beta), static_cast<double>(x)
+    );
+}
+
+XSF_HOST_DEVICE inline std::complex<double> eval_jacobi(double n, double alpha, double beta, std::complex<double> x) {
+    return detail::eval_jacobi_impl(n, alpha, beta, x);
+}
+
```
j-bowhay, line 78: What about `std::complex<float> eval_jacobi`?

include/xsf/orthogonal_eval.h
```diff
@@ -0,0 +1,111 @@
...
+XSF_HOST_DEVICE inline double eval_jacobi_l(std::ptrdiff_t n, double alpha, double beta, double x) {
```
j-bowhay: I think we would want this to be another overload

j-bowhay: Also possibly, the first argument should be long?

include/xsf/orthogonal_eval.h
```diff
@@ -0,0 +1,111 @@
...
+            t = 2.0 * k + alpha + beta;
+            d = ((t * (t + 1.0) * (t + 2.0)) * (x - 1.0) * p + 2.0 * k * (k + beta) * (t + 2.0) * d) /
+                (2.0 * (k + alpha + 1.0) * (k + alpha + beta + 1.0) * t);
+            p = d + p;
+        }
+        return binom(n + alpha, n) * p;
+    }
+}
+
+XSF_HOST_DEVICE inline float eval_jacobi_l(std::ptrdiff_t n, float alpha, float beta, float x) {
```
j-bowhay: Same comment about making overload

include/xsf/orthogonal_eval.h
khushi-411: Hi, @j-bowhay! Thanks for reviewing the PR.

Added the `std::complex<float>` overload.

include/xsf/orthogonal_eval.h
khushi-411: Nice catch, thanks, @j-bowhay! updated the PR.

tests/xsf_tests/test_orthogonal_eval.cpp
```diff
@@ -0,0 +1,329 @@
...
+}
+
+template <typename T>
+T polyval(const std::vector<double> &coeffs, T x) {
+    T out = 0.0;
+    for (auto it = coeffs.rbegin(); it != coeffs.rend(); ++it) {
+        out = out * x + *it;
+    }
+    return out;
+}
```
j-bowhay: https://github.com/scipy/xsf/blob/5759f206993cc2c34301e13ecb656446d5e79e45/include/xsf/cephes/polevl.h#L70

include/xsf/orthogonal_eval.h
dschmitz89: Puh, it's a pity that there are no tests with complex inputs in SciPy currently. Then let's juse use the current SciPy outputs as a reference so that we at least do not make things worse. Practically: please add a handful of tests with complex inputs that verify that we get the same results as SciPy.

khushi-411: Sounds good, thanks! I've added tests for `std::complex<double>` and `std::complex<float>` inputs using hardcoded SciPy outputs as the reference. I also added a polynomial based test for complex inputs.

tests/xsf_tests/test_orthogonal_eval.cpp
khushi-411: Updated, thanks!

khushi-411:
@dschmitz89, @j-bowhay, I was wondering why we add another overload for `float` inputs and then `static_cast` them to use the `double` functions, instead of simply using templates. The `static_cast` may cause precision issues; in fact, I noticed multiple CI warnings about this.

Although I do agree, templating could unintentionally expand the set of supported dtype combinations, which may not be needed.

include/xsf/orthogonal_eval.h
khushi-411: @j-bowhay, wouldn't `std::ptrdiff_t` be a better choice than `long`? On 64-bit systems `std::ptrdiff_t` represents 64 bits whereas `long` is still 32-bit but for 32-bit systems both represents 32-bits.

Reference: https://github.com/scipy/xsf/pull/53#issuecomment-3122106603

What do you think, should I replace it with `std::ptrdiff_t` or follow the current codebase for now?

j-bowhay: Fair point you are probably right, feel free to ignore

j-bowhay:
> @dschmitz89, @j-bowhay, I was wondering why we add another overload for `float` inputs and then `static_cast` them to use the `double` functions, instead of simply using templates. The `static_cast` may cause precision issues; in fact, I noticed multiple CI warnings about this.


The computation generally needs to be done in double precision for sufficient accuracy

dschmitz89:
Could you also add permalinks to the tests as well here?

I am wondering why there is a lot of machinery in the tests. Is it possible to simplify those in a follow up?

khushi-411:
Thanks, @dschmitz89! I've added permalinks to the tests. I'll also open a follow-up PR to simplify the tests.

khushi-411:
Gentle ping, @dschmitz89, @j-bowhay! The CI is green, and all review comments have been addressed.

dschmitz89:
This LGTM, let's wait for a final approval by @j-bowhay .

khushi-411:
Hi, gentle ping, @j-bowhay!

khushi-411:
Gentle ping, @j-bowhay!

khushi-411:
Thanks so much for reviewing the PR, @dschmitz89, @j-bowhay!

PR #211, opened 2026-07-10
ENH: add `iv_ratioinv`

dschmitz89:
#### Reference issue
Towards https://github.com/scipy/scipy/issues/20253

This PR implements the new function `iv_ratioinv` which solves $r=I_v(x)/I_{v-1}(x)$ for x given $r$ and $v$. This equation occurs in MLE of distributions such as von Mises-Fisher. Currently, an adhoc solution exists in SciPy.

#### Additional information
The function is a simple derivative free root finding procedure. Tight brackets for the root were taken from
[this paper](https://pubs.ams.org/journals/mcom/1974-28-125/S0025-5718-1974-0333287-7/S0025-5718-1974-0333287-7.pdf?utm_source=consensus). In case those bounds fail, we fall back to the bracketing routine for monotonic functions. Once, bounds are found we run Chandrupatla's algorithm to find the exact root.

The SciPy issue contains a gradient based approach but I found this one easier to implement and assume that it is more robust as the objective can be near flat in the tails.

#### AI Generation Disclosure
LLMs helped inverting the bounds equations.

### DIFF
--- a/include/xsf/iv_ratio.h
+++ b/include/xsf/iv_ratio.h
@@ -1,4 +1,4 @@
-// Numerically stable computation of iv(v+1, x) / iv(v, x)
+// Numerically stable computation of iv(v+1, x) / iv(v, x) and its inverse
 
 #pragma once
 
@@ -164,6 +164,94 @@ XSF_HOST_DEVICE inline double iv_ratio_c(double v, double x) {
 
 XSF_HOST_DEVICE inline float iv_ratio_c(float v, float x) {
     return iv_ratio_c(static_cast<double>(v), static_cast<double>(x));
+}
+
+XSF_HOST_DEVICE inline double iv_ratioinv(double v, double r) {
+    if (std::isnan(v) || std::isnan(r)) {
+        return std::numeric_limits<double>::quiet_NaN();
+    }
+    if (v < 0.5 || r <= 0. || r >= 1) {
+        // iv_ratioinv is only defined for v >= 0.5
+        // for r=0 or r=1, the inverse is not unique (x=0 and x=inf respectively)
+        set_error("iv_ratioinv", SF_ERROR_DOMAIN, NULL);
+        return std::numeric_limits<double>::quiet_NaN();
+    }
+    // Algorithm description: use Chandrupatla's method to find the root of
+    // f(x) = iv_ratio(v, x) - r (or f(x) = 1 - iv_ratio_c(v, x) - r if r > 0.5).
+    // Bounds from Amos, 1973. "Computation of Modified Bessel Function And Their Ratios".
+    // That paper defines the ratio as iv(v+1, x)/iv(v, x), but we define it as iv(v, x)/iv(v-1, x).
+    // Substitute v -> v-1 in eq. (9) and eq. (16) yields the following bounds for x:
+    //   x/(v-1/2 + sqrt(x^2 + (v+1/2)^2)) <= r <= x/(v-1 + sqrt(x^2 + (v+1)^2))
+    //
+    // Inverting (inequality direction flips):
+    //   lower x (from upper r-bound): r*[(v-1) + sqrt((v-1)^2 + 4v*(1-r^2))] / (1-r^2)
+    //   upper x (from lower r-bound): r*[(v-0.5) + sqrt((v-0.5)^2 + 2v*(1-r^2))] / (1-r^2)
+
+    double r_c = 1.0 - r;
+    double one_minus_r_sq = r_c * (1.0 + r); // 1 - r^2
+
+    double vm1 = v - 1.0;
+    double lower_bound = r * (vm1 + std::sqrt(vm1 * vm1 + 4.0 * v * one_minus_r_sq)) / one_minus_r_sq;
+
+    double vm05 = v - 0.5;
+    double upper_bound = r * (vm05 + std::sqrt(vm05 * vm05 + 2.0 * v * one_minus_r_sq)) / one_minus_r_sq;
+
+    // For small r both bounds converge to 2v*r (== true x). Ensure bracket has width.
+    if (upper_bound <= lower_bound) {
+        upper_bound = 2.0 * lower_bound;
+    }
+
+    auto func = [v, r, r_c](double x) {
+        if (r <= 0.5) {
+            return iv_ratio(v, x) - r;
+        }
+        return r_c - iv_ratio_c(v, x);
+    };
+
+    double xl = lower_bound;
+    double xr = upper_bound;
+    double f_xl = func(xl);
+    double f_xr = func(xr);
+
+    if (f_xl * f_xr > 0) {
+        // The Amos bounds are theoretically correct but may yield an invalid bracket
+        // due to precision issues in iv_ratio or iv_ratio_c. This happens especially for very small or large r.
+        // Fallback: use bracket_root_for_cdf_inversion to find a valid bracket.
+        // Bracketing parameters taken from gdtrib, only difference: our function is increasing
+        auto [b_xl, b_xr, b_f_xl, b_f_xr, bracket_status] = detail::bracket_root_for_cdf_inversion(
+            func, 1.0, std::numeric_limits<double>::min(), std::numeric_limits<double>::max(), -0.875, 7.0, 0.125, 8,
+            true, 342
+        );
+        if (bracket_status == 1) {
+            set_error("iv_ratioinv", SF_ERROR_UNDERFLOW, NULL);
+            return 0.0;
+        }
+        if (bracket_status == 2) {
+            set_error("iv_ratioinv", SF_ERROR_OVERFLOW, NULL);
+            return std::numeric_limits<double>::infinity();
+        }
+        if (bracket_status >= 3) {
+            set_error("iv_ratioinv", SF_ERROR_OTHER, "Computational Error");
+            return std::numeric_limits<double>::quiet_NaN();
+        }
+        xl = b_xl;
+        xr = b_xr;
+        f_xl = b_f_xl;
+        f_xr = b_f_xr;
+    }
+
+    auto [result, root_status] =
+        detail::find_root_chandrupatla(func, xl, xr, f_xl, f_xr, std::numeric_limits<double>::epsilon(), 1e-308, 1000);
+    if (root_status) {
+        // Root finding failed. This should never happen.
+        set_error("iv_ratioinv", SF_ERROR_OTHER, "Computational Error");
+        return std::numeric_limits<double>::quiet_NaN();
+    }
+    return result;
+}
+
+XSF_HOST_DEVICE inline float iv_ratioinv(float v, float r) {
+    return static_cast<float>(iv_ratioinv(static_cast<double>(v), static_cast<double>(r)));
 }
 
 } // namespace xsf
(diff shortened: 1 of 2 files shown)


dschmitz89:
Cc @fancidev in case you want to take a look

include/xsf/iv_ratio.h
```diff
@@ -1,4 +1,4 @@
-// Numerically stable computation of iv(v+1, x) / iv(v, x)
+// Numerically stable computation of iv(v+1, x) / iv(v, x) and its inverse
```
dschmitz89:
```suggestion
// Numerically stable computation of iv(v, x) / iv(v - 1, x) and its inverse
```

include/xsf/iv_ratio.h
```diff
@@ -166,4 +166,92 @@ XSF_HOST_DEVICE inline float iv_ratio_c(float v, float x) {
...
+    double xr = upper_bound;
+    double f_xl = func(xl);
+    double f_xr = func(xr);
+
+    if (f_xl * f_xr > 0) {
+        // The Amos bounds are theoretically correct but may yield an invalid bracket
+        // due to precision issues in iv_ratio or iv_ratio_c. This happens especially for very small or large r.
+        // Fallback: use bracket_root_for_cdf_inversion to find a valid bracket.
+        // Bracketing parameters taken from gdtrib, only difference: our function is increasing
+        auto [b_xl, b_xr, b_f_xl, b_f_xr, bracket_status] = detail::bracket_root_for_cdf_inversion(
```
dschmitz89, line 233: This function should be renamed as it works with any monotonic function, not just CDFs.

dschmitz89:
@fbourgey Would you have time to review?

include/xsf/iv_ratio.h
```diff
@@ -166,4 +166,92 @@ XSF_HOST_DEVICE inline float iv_ratio_c(float v, float x) {
...
+XSF_HOST_DEVICE inline double iv_ratioinv(double v, double r) {
+    if (std::isnan(v) || std::isnan(r)) {
+        return std::numeric_limits<double>::quiet_NaN();
+    }
+    if (v < 0.5 || r <= 0. || r >= 1) {
+        // iv_ratioinv is only defined for v >= 0.5
+        // for r=0 or r=1, the inverse is not unique (x=0 and x=inf respectively)
+        set_error("iv_ratioinv", SF_ERROR_DOMAIN, NULL);
+        return std::numeric_limits<double>::quiet_NaN();
+    }
```
fbourgey: I am probably missing something but why is the inverse not unique in that case?

what's wrong with?


```suggestion
   if (!std::isfinite(v) || v < 0.5 || r < 0.0 || r > 1.0) {
        // iv_ratioinv is only defined for v >= 0.5
        set_error("iv_ratioinv", SF_ERROR_DOMAIN, NULL);
        return std::numeric_limits<double>::quiet_NaN();
    }
    if (r == 0.0) {
        return 0.0;
    }
    if (r == 1.0) {
        return std::numeric_limits<double>::infinity();
    }
```

If so, we could add some tests for $r=0$ and $r=1$

fbourgey (commented):
When $v=\frac12$, it appears we have $I_{1/2}(x)/I_{-1/2}(x)=\tanh(x) = r$, hence $x = \text{atanh}(r)$. 

Could you add some tests for $v=0.5$, ideally for the extreme tails? 

I wonder if `std::nextafter(1.0, 0.0)` and `std::nextafter(0.0, 1.0)` pass?

dschmitz89:
> When v = 1 2 , it appears we have I 1 / 2 ( x ) / I − 1 / 2 ( x ) = tanh ⁡ ( x ) = r , hence x = atanh ( r ) .
> 
> Could you add some tests for v = 0.5 , ideally for the extreme tails?
> 
> I wonder if `std::nextafter(1.0, 0.0)` and `std::nextafter(0.0, 1.0)` pass?

They do not unfortunately pass a roundtrip test. I think it is not possible in double precision. Could you take one last look?

Edit: the `iv_ratio` function itself does not use the analytical method yet, will try that out.

include/xsf/iv_ratio.h
```diff
@@ -157,13 +161,117 @@ XSF_HOST_DEVICE inline double iv_ratio_c(double v, double x) {
...
+    if (v == 0.5) {
+        // Closed-form solution for v = 0.5: iv_ratio(0.5, x) = tanh(x)
+        // Since tanh(x) = 2*expit(2*x) - 1 = r, we have expit(2*x) = (1+r)/2
+        // For r > 0.5, use logit((1+r)/2) for better stability near r=1
+        // For r <= 0.5, atanh is sufficiently stable
+        if (r > 0.5) {
+            return 0.5 * logit((1.0 + r) * 0.5);
+        } else {
+            return std::atanh(r);
+        }
```
fbourgey:
```suggestion
    if (v == 0.5) {
      return std::atanh(r);
    }
```

would not be enough? 

On my machine os-arm64, `iv_ratioinv(0.5, nextafter(1.0, 0.0))` gives `inf` while it should be `18.714973875118524`. I think this is because (1+r)*0.5 rounds to 1 in that case.

dschmitz89:
@fbourgey I added the tests for the `v=0.5` case you asked for. I think this should be good now.

tests/xsf_tests/test_iv_ratioinv.cpp
```diff
@@ -0,0 +1,100 @@
...
+    const double nan = std::numeric_limits<double>::quiet_NaN();
+    REQUIRE(std::isnan(xsf::iv_ratioinv(nan, 0.5)));
+    REQUIRE(std::isnan(xsf::iv_ratioinv(1.0, nan)));
+}
+
+TEST_CASE("iv_ratioinv domain error", "[iv_ratioinv][xsf_tests]") {
+    const double inf = std::numeric_limits<double>::infinity();
+    REQUIRE(std::isnan(xsf::iv_ratioinv(0.4, 0.5)));  // v < 0.5
+    REQUIRE(std::isnan(xsf::iv_ratioinv(1.0, -1.0))); // y < 0
+    REQUIRE(std::isnan(xsf::iv_ratioinv(1.0, 2.0)));  // y > 1
```
fbourgey: to match the function argument

```suggestion
    REQUIRE(std::isnan(xsf::iv_ratioinv(1.0, -1.0))); // r < 0
    REQUIRE(std::isnan(xsf::iv_ratioinv(1.0, 2.0)));  // r > 1
```

tests/xsf_tests/test_iv_ratioinv.cpp
```diff
@@ -0,0 +1,100 @@
...
+    REQUIRE(std::isnan(xsf::iv_ratioinv(inf, 0.5)));  // v = inf
+}
+
+TEST_CASE("iv_ratioinv domain boundary", "[iv_ratioinv][xsf_tests]") {
+    const double inf = std::numeric_limits<double>::infinity();
+    REQUIRE(xsf::iv_ratioinv(1.0, 0.0) == 0.0); // y = 0
+    REQUIRE(xsf::iv_ratioinv(1.0, 1.0) == inf); // y = 1
```
fbourgey:
```suggestion
    REQUIRE(xsf::iv_ratioinv(1.0, 0.0) == 0.0); // r = 0
    REQUIRE(xsf::iv_ratioinv(1.0, 1.0) == inf); // r = 1
```

tests/xsf_tests/test_iv_ratioinv.cpp
```diff
@@ -0,0 +1,100 @@
+#include "../testing_utils.h"
+#include "xsf/config.h"
+#include <xsf/iv_ratio.h>
+
+TEST_CASE("iv_ratioinv round-trip single-double", "[iv_ratioinv][xsf_tests]") {
```
fbourgey: a more detailed test description

```suggestion
TEST_CASE("iv_ratioinv round trip in single and double precision", "[iv_ratioinv][xsf_tests]") {
```

include/xsf/iv_ratio.h
fbourgey: I'd use `// There is no unique limit as both v and x tend to infinity.`

include/xsf/iv_ratio.h
```diff
@@ -157,13 +161,109 @@ XSF_HOST_DEVICE inline double iv_ratio_c(double v, double x) {
...
+    }
+    if (r == 1.0) {
+        return std::numeric_limits<double>::infinity();
+    }
+    if (v == 0.5) {
+        return std::atanh(r);
+    }
+    // Algorithm description: use Chandrupatla's method to find the root of
+    // f(x) = iv_ratio(v, x) - r (or f(x) = 1 - iv_ratio_c(v, x) - r if r > 0.5).
+    // Bounds from Amos, 1973. "Computation of Modified Bessel Function And Their Ratios".
```
fbourgey: From https://www.jstor.org/stable/2005830, the paper is from 1974.

```suggestion
    // Bounds from Amos (1974), "Computation of Modified Bessel Functions and Their Ratios".
```

include/xsf/iv_ratio.h
```diff
@@ -157,13 +161,109 @@ XSF_HOST_DEVICE inline double iv_ratio_c(double v, double x) {
...
+    // That paper defines the ratio as iv(v+1, x)/iv(v, x), but we define it as iv(v, x)/iv(v-1, x).
+    // Substitute v -> v-1 in eq. (9) and eq. (16) yields the following bounds for x:
```
fbourgey:
```suggestion
    // Substituting v -> v-1 in eq. (9) and eq. (16) yields the following bounds for x:
```

include/xsf/iv_ratio.h
```diff
@@ -157,13 +161,109 @@ XSF_HOST_DEVICE inline double iv_ratio_c(double v, double x) {
...
+        // Equivalently: iv_ratio_c(0.5, x) = 1 - tanh(x) = 2 * expit(-2*x)
+        return 2 * expit(-2 * x);
     }
 }
 
 XSF_HOST_DEVICE inline float iv_ratio_c(float v, float x) {
     return iv_ratio_c(static_cast<double>(v), static_cast<double>(x));
 }
 
+XSF_HOST_DEVICE inline double iv_ratioinv(double v, double r) {
```
fbourgey, line 176: I think a short description would be useful.

```suggestion
// Compute x such that I_v(x) / I_{v-1}(x) = r for v >= 0.5 and 0 <= r <= 1.
XSF_HOST_DEVICE inline double iv_ratioinv(double v, double r) {
```

include/xsf/iv_ratio.h
fbourgey, line 120: I would add a short description too

`// Compute 1 - I_v(x) / I_{v-1}(x) directly to preserve accuracy near 1.`

fbourgey, line 80: `// Compute I_v(x) / I_{v-1}(x) for v >= 0.5 and x >= 0.`

fbourgey (commented):
@dschmitz89 I think this is ready to merge. I'd say go ahead and self-merge if OK with you. I added some (minor) suggestions. I let you decide if you want to add them, notably the descriptions.

fbourgey:
I went ahead and updated the comments. Thanks @dschmitz89.

PR #212, opened 2026-07-10
ENH: add Newtonian root finder to root finding tools

dschmitz89:
#### Reference issue
Follow up to #130

#130 implemented the inverse digamma function using a Newtonian root finder. I am sure that a general Newtonian root finder will be useful in future, so I added a reusable method for it. `digammainv` is then updated to use this method.

#### Additional information
The technicalities mostly follow what I saw in `tools.h`. 

#### AI Generation Disclosure
AI helped with the implementation and discussions about default parameters.

### DIFF
--- a/include/xsf/tools.h
+++ b/include/xsf/tools.h
@@ -422,5 +422,29 @@ namespace detail {
         return {std::numeric_limits<double>::quiet_NaN(), 1};
     }
 
+    /* Find root of a scalar function using Newton-Raphson.
+     *
+     * Function func must return a pair (f(x), f'(x)).
+     */
+    template <typename Function>
+    XSF_HOST_DEVICE inline std::pair<double, int> find_root_newton(
+        Function func, double x, double rtol = 4 * std::numeric_limits<double>::epsilon(), double atol = 0.0,
+        std::uint64_t maxiter = 100
+    ) {
+        for (std::uint64_t i = 0; i < maxiter; i++) {
+            auto [f, df] = func(x);
+            if (f == 0.0) {
+                return {x, 0};
+            }
+            double step = f / df;
+            double x_next = x - step;
+            if (x == x_next || std::abs(step) <= rtol * std::abs(x) + atol) {
+                return {x_next, 0};
+            }
+            x = x_next;
+        }
+        return {x, 1};
+    }
+
 } // namespace detail
 } // namespace xsf
--- a/include/xsf/tools.h
+++ b/include/xsf/tools.h
@@ -426,42 +426,52 @@ namespace detail {
      *
      * Function func must return a pair (f(x), f'(x)).
      */
+
+    enum class NewtonRootFinderStatus {
+        SUCCESS = 0,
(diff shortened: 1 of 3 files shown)


include/xsf/tools.h
```diff
@@ -422,5 +422,29 @@ namespace detail {
...
+     * Function func must return a pair (f(x), f'(x)).
+     */
+    template <typename Function>
+    XSF_HOST_DEVICE inline std::pair<double, int> find_root_newton(
+        Function func, double x, double rtol = 4 * std::numeric_limits<double>::epsilon(), double atol = 0.0,
+        std::uint64_t maxiter = 100
+    ) {
+        for (std::uint64_t i = 0; i < maxiter; i++) {
+            auto [f, df] = func(x);
+            if (f == 0.0) {
```
fbourgey, line 455: can we add guardrails for non-finite f/df, zero derivative, non-finite iterate? I think we can have cases where the return status is 0 without finding a root.

include/xsf/tools.h
```diff
@@ -422,5 +422,47 @@ namespace detail {
...
+            }
+            if (std::isnan(f) || std::isnan(df)) {
+                return {std::numeric_limits<double>::quiet_NaN(), 3};
+            }
+            if (std::isinf(f) && (f > 0)) {
+                return {std::numeric_limits<double>::infinity(), 4};
+            }
+            if (std::isinf(f) && (f < 0)) {
+                return {-std::numeric_limits<double>::infinity(), 4};
+            }
```
fbourgey, line 463: If f=0, then we return x directly. Also, if f is inf, why would the root be inf? Maybe something like the below is enough?

```suggestion
            auto [f, df] = func(x);
            if (f == 0.0) {
                return {x, 0};
            }
            if (!std::isfinite(f) || !std::isfinite(df)) {
                return {std::numeric_limits<double>::quiet_NaN(), 3};
            }
            if (df == 0.0) {
                return {std::numeric_limits<double>::quiet_NaN(), 2};
            }
```

include/xsf/tools.h
```diff
@@ -422,5 +422,47 @@ namespace detail {
...
+            double step = f / df;
+            double x_next = x - step;
+            if (x == x_next || std::abs(step) <= rtol * std::abs(x) + atol) {
+                return {x_next, 0};
+            }
```
fbourgey, line 471: we can check that x_next is still finite. We could also do it for step but I think this is redundant given x_next = x-step.

```suggestion
            double x_next = x - step;
            if (!std::isfinite(x_next)) {
                return {std::numeric_limits<double>::quiet_NaN(), 3};
            }
            if (x == x_next || std::abs(step) <= rtol * std::abs(x) + atol) {
                return {x_next, 0};
            }
```

fbourgey (commented):
Some questions/comments. Also, should we need a enum for the different status? or at least add a comment for those.

fbourgey:
Thanks @dschmitz89!

PR #213, opened 2026-07-12
MAINT: migrate stirling2 from scipy to xsf

dschmitz89:
#### Reference issue
Towards #170

### DIFF
--- a/include/xsf/stirling2.h
+++ b/include/xsf/stirling2.h
@@ -1,0 +1,125 @@
+/* Stirling numbers of the second kind
+ *
+ * SYNOPSIS: Stirling numbers of the second kind count the number of ways to
+ *  make a partition of n distinct elements into k non-empty subsets.
+ *
+ * DESCRIPTION: n is the number of distinct elements of the set to be
+ *  partitioned and k is the number of non-empty subsets. Values for n < 0 or
+ *  k < 0 are interpreted as 0.
+ *
+ * NOTE: Original implementation by Lucas Roberts. Ported to xsf by SciPy
+ *  developers.
+ *
+ * References:
+ *  - N. M. Temme, "Asymptotic estimates of Stirling numbers", Studies in
+ *    Applied Mathematics, 89(3):233-243, 1993.
+ */
+
+#pragma once
+
+#include "binom.h"
+#include "config.h"
+#include "error.h"
+#include "lambertw.h"
+
+namespace xsf {
+namespace detail {
+
+    // Dynamic programming method for small n.
+    // Only valid for n <= 50; the stack array is sized accordingly (max 25 elements).
+    XSF_HOST_DEVICE inline double stirling2_dp(double n, double k) {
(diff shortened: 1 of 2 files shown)


include/xsf/stirling2.h
```diff
@@ -0,0 +1,122 @@
...
+} // namespace detail
+
+// Stirling numbers of the second kind S(n, k):
+// Uses dynamic programming for n <= 50, Temme approximation otherwise.
+XSF_HOST_DEVICE inline double stirling2(double n, double k) {
+    if (n <= 50) {
+        return detail::stirling2_dp(n, k);
+    }
+    return detail::stirling2_temme(n, k);
+}
```
j-bowhay: float overload is also needed

include/xsf/stirling2.h
```diff
@@ -0,0 +1,132 @@
...
+    }
+} // namespace detail
+
+XSF_HOST_DEVICE inline double stirling2(double n, double k) { return detail::stirling2(n, k); }
+
+XSF_HOST_DEVICE inline float stirling2(float n, float k) {
+    return static_cast<float>(detail::stirling2(static_cast<double>(n), static_cast<double>(k)));
+}
+
+template <typename T1, typename T2>
```
dschmitz89: This overload was added to make tests with integer inputs pass. We might also simply switch to doubles in tests instead. The `is_integral` stuff will likely not work on GPU.

j-bowhay: I would change the tests rather than make the code more complicated

tests/xsf_tests/test_stirling2.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+
+TEST_CASE("stirling2 k out of range", "[stirling2][xsf_tests]") {
+    // k > n or k = 0 with n > 0 gives 0
+    REQUIRE(xsf::stirling2(3.0, 4.0) == 0.0);
+    REQUIRE(xsf::stirling2(5.0, 0.0) == 0.0);
+    REQUIRE(xsf::stirling2(0.0, 1.0) == 0.0);
+}
+
+TEST_CASE("stirling2 temme approximation accuracy", "[stirling2][xsf_tests]") {
```
fbourgey, line 84: AFAICT, we check values for $k=1, n, n-1$, and $2$ so this is not testing eg the interior case S(75,30). So, should we add a test for k=1..n similar to what SciPy does [here](https://github.com/scipy/scipy/blob/main/scipy/special/tests/test_basic.py#L4883-L4891)? I think this is where the Temme approximation does the real work.

dschmitz89: The test compares the `exact` code path with the `inexact` one. The `exact` path is still implemented in Python as it requires python integers from what I understand: see [here](https://github.com/scipy/scipy/blob/20a186fb9611ff4787f0fd6d8784d42339fdcd27/scipy/special/_basic.py#L3443). Unless we migrate that code path also to xsf that test has to stay in SciPy only.

PR #217, opened 2026-07-15
MAINT: remove remaining ufl variable in amos

fbourgey:
CI is failing on https://github.com/scipy/scipy/pull/25636. It seems we forgot to remove the `ufl` variable in https://github.com/scipy/xsf/pull/132

#### AI Generation Disclosure

I asked Codex for any remaining missing cleanups

### DIFF
--- a/include/xsf/amos/amos.h
+++ b/include/xsf/amos/amos.h
@@ -1286,8 +1286,8 @@ namespace amos {
         //***END PROLOGUE  ZBESH
 
         std::complex<double> zn, zt, csgn;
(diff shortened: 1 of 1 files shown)


dschmitz89 (approved):

dschmitz89:
We should add a CI job with `-Werror` to xsf. @lucascolley Do we need a new pixi task definition for that? I am starting to like pixi, would use this as an opportunity to get more familiar with it.

lucascolley:
> We should add a CI job with `-Werror` to xsf. @lucascolley Do we need a new pixi task definition for that? I am starting to like pixi, would use this as an opportunity to get more familiar with it.

Sounds reasonable! @steppi maybe we can just add it to the existing tasks?

PR #218, opened 2026-07-16
ENH: improve iv for tiny inputs

dschmitz89:
#### Reference issue
Towards https://github.com/scipy/scipy/issues/25287

This PR fixes `iv` for small inputs by using the leading order of the series. The same is already done for `jv` in [this section](https://github.com/scipy/xsf/blob/6dacc318b2028e193ccbed3962704d21cf2b2442/include/xsf/cephes/jv.h#L609). By fixing `iv` we fix `spherical_i` and thereby the mentioned SciPy issue.

#### AI Generation Disclosure
LLMs created the fix, tests are mine.

### DIFF
--- a/tests/xsf_tests/test_sph_bessel.cpp
+++ b/tests/xsf_tests/test_sph_bessel.cpp
@@ -152,3 +152,21 @@ TEST_CASE("spherical_y_jac reflection derivative", "[bessel][xsf_tests]") {
     CAPTURE(n, z, result_spherical_y_jac, ref_spherical_y_jac, rel_err_spherical_y_jac, rtol);
     REQUIRE(rel_err_spherical_y_jac <= rtol);
 }
+
+TEST_CASE("spherical_i tiny inputs", "[bessel][xsf_tests]") {
+    using test_case = std::tuple<long, double, double, double>;
+
+    // Reference values computed with mpmath.
+    auto [n, z, ref, rtol] = GENERATE(
+        test_case{1, 1e-50, 3.333333333333333e-51, 1e-14}, test_case{100, 0.1, 7.463271153267418e-290, 1e-10},
+        test_case{10, 5.2250558491838786e-27, 1.10313444760931e-273, 1e-14},
+        test_case{3, 1e-50, 9.523809523809524e-153, 1e-14}, test_case{10, 1e-29, 7.273091945557419e-301, 1e-8},
+        test_case{20, 1e-14, 7.625979004892142e-306, 1e-8}, test_case{200, 5, 3.168106195219311e-297, 1e-10}
+    );
+
+    double result = xsf::sph_bessel_i(n, z);
+    double rel_err = xsf::extended_relative_error(result, ref);
+
+    CAPTURE(n, z, result, ref, rel_err, rtol);
+    REQUIRE(rel_err <= rtol);
+}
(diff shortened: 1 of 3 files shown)


dschmitz89:
CC @WarrenWeckesser

dschmitz89:
Ouch, turned out I only ran the new test I added locally.

dschmitz89:
@gertingold This PR also covers the modified spherical besssel function, in particular it's behavior for very tiny inputs. Since you are more familiar with them than many of us, a review would be highly appreciated.

gertingold:
As the leading term of the Taylor series of $I_\nu(z)$ and $J_\nu(z)$ agree, it seems fine to simply use the code from `cephes/jv.h`. My only suggestion would be to consider splitting the tests for the Bessel functions and for the spherical Bessel functions. The change in `cephes/scipy_iv.h` addresses the modified Bessel function so that there should be tests of this function in `test_bessel.cpp`. Presently, the proposed tests verify the code via the modified spherical Bessel function only in an indirect way. Having tests of the modified spherical Bessel function is certainly good, but they could go to `test_sph_bessel.cpp` which was introduced in the context of #221. `test_bessel.cpp` could contain corresponding direct tests for the modified Bessel function.

I am suggesting this split because in view of the notation $i_\nu(z)$ for the modified spherical Bessel function, it is easy to mistake `iv` for the modified spherical Bessel function. When I first looked at the code, I compared with the Taylor series of the modified spherical Bessel function while the code of course represents the first term in the Taylor series of the modified Bessel function.

dschmitz89:
> As the leading term of the Taylor series of I ν ( z ) and J ν ( z ) agree, it seems fine to simply use the code from `cephes/jv.h`. My only suggestion would be to consider splitting the tests for the Bessel functions and for the spherical Bessel functions. The change in `cephes/scipy_iv.h` addresses the modified Bessel function so that there should be tests of this function in `test_bessel.cpp`. Presently, the proposed tests verify the code via the modified spherical Bessel function only in an indirect way. Having tests of the modified spherical Bessel function is certainly good, but they could go to `test_sph_bessel.cpp` which was introduced in the context of #221. `test_bessel.cpp` could contain corresponding direct tests for the modified Bessel function.
> 
> I am suggesting this split because in view of the notation i ν ( z ) for the modified spherical Bessel function, it is easy to mistake `iv` for the modified spherical Bessel function. When I first looked at the code, I compared with the Taylor series of the modified spherical Bessel function while the code of course represents the first term in the Taylor series of the modified Bessel function.

Thanks for the review @gertingold , I applied the test split you suggested.

lucascolley:
there is a merge conflict

tests/xsf_tests/test_sph_bessel.cpp
```diff
@@ -152,3 +152,21 @@ TEST_CASE("spherical_y_jac reflection derivative", "[bessel][xsf_tests]") {
     CAPTURE(n, z, result_spherical_y_jac, ref_spherical_y_jac, rel_err_spherical_y_jac, rtol);
     REQUIRE(rel_err_spherical_y_jac <= rtol);
 }
+
+TEST_CASE("spherical_i tiny inputs", "[bessel][xsf_tests]") {
```
lucascolley: should this tag say `sph_bessel` instead of `bessel` everywhere in this file?

dschmitz89: Good suggestion. Done.

dschmitz89:
Could you take a last look @lucascolley ? I would like to avoid getting merge conflicts in this PR again.

lucascolley (approved):

PR #219, opened 2026-07-18
ENH: add integer-parameter signature in numpy.h

fbourgey:
Following https://github.com/scipy/xsf/pull/209, I was trying to use Jacobi and Shifted Jacobi polynomials from XSF. It appears those ufunc signatures will be useful for `scipy/special/_special_ufuncs.cpp` for Jacobi and the following polynomials.

#### AI Generation Disclosure

Codex reviewed.

### DIFF
--- a/include/xsf/numpy.h
+++ b/include/xsf/numpy.h
@@ -190,6 +190,7 @@ namespace numpy {
     using D_DDDD = void (*)(cdouble, cdouble &, cdouble &, cdouble &, cdouble &);
 
     // 2 inputs, 1 output
+    using pd_d = double (*)(std::ptrdiff_t, double);
     using qf_f = float (*)(long long int, float);
     using qd_d = double (*)(long long int, double);
     using ff_f = float (*)(float, float);
@@ -236,6 +237,7 @@ namespace numpy {
     using dd_dddd = void (*)(double, double, double &, double &, double &, double &);
 
     // 3 inputs, 1 output
+    using pdd_d = double (*)(std::ptrdiff_t, double, double);
     using fff_f = float (*)(float, float, float);
     using ddd_d = double (*)(double, double, double);
     using flf_f = double (*)(double, long, double);
@@ -268,6 +270,7 @@ namespace numpy {
     using qqd_ddd = void (*)(long long int, long long int, double, double &, double &, double &);
 
     // 4 inputs, 1 output
+    using pddd_d = double (*)(std::ptrdiff_t, double, double, double);
     using llld_d = double (*)(long int, long int, long int, double);
     using qqqd_d = double (*)(long long int, long long int, long long int, double);
     using qqqF_F = cfloat (*)(long long int, long long int, long long int, cfloat);


dschmitz89:
Ay, didn't expect this. @khushi-411 Was there a hard requirement why we need `std::ptrdiff_t` instead of a regular int for the Jacobi functions?

khushi-411:
Hi, @dschmitz89, @fbourgey! The main reason was that the windows tests were failing in CI when I used a test case with an input value of dtype `long` so I used `std::ptrdiff_t`.

From here: https://github.com/scipy/xsf/pull/209#discussion_r3529288645

> On 64-bit systems std::ptrdiff_t represents 64 bits whereas long is still 32-bit but for 32-bit systems both represents 32-bits.
>
> Reference: https://github.com/scipy/xsf/pull/53#issuecomment-3122106603

dschmitz89:
> Hi, @dschmitz89, @fbourgey! The main reason was that the windows tests were failing in CI when I used a test case with an input value of dtype `long` so I used `std::ptrdiff_t`.
> 
> From here: https://github.com/scipy/xsf/pull/209#discussion_r3529288645
> 
> > On 64-bit systems std::ptrdiff_t represents 64 bits whereas long is still 32-bit but for 32-bit systems both represents 32-bits.
> >
> > Reference: https://github.com/scipy/xsf/pull/53#issuecomment-3122106603
> 
> 

Thanks for the explanation. To me this looks good then but it would be great if @steppi could take a final look.

fbourgey:
> > Hi, @dschmitz89, @fbourgey! The main reason was that the windows tests were failing in CI when I used a test case with an input value of dtype `long` so I used `std::ptrdiff_t`.
> > From here: [#209 (comment)](https://github.com/scipy/xsf/pull/209#discussion_r3529288645)
> > > On 64-bit systems std::ptrdiff_t represents 64 bits whereas long is still 32-bit but for 32-bit systems both represents 32-bits.
> > > Reference: [#53 (comment)](https://github.com/scipy/xsf/pull/53#issuecomment-3122106603)
> 
> Thanks for the explanation. To me this looks good then but it would be great if @steppi could take a final look.

@steppi WDYT?

fbourgey:
Why not simply use `int` instead of `std::ptrdiff_t`? The following https://github.com/fbourgey/xsf/commit/481410774c994ae986836fa4725c0d9fca0f92a9 commit seems to pass CI.

khushi-411:
@fbourgey, I thought since `int` is always 32-bit on both 32-bit and 64-bit systems, so I thought `std::ptrdiff_t` would be a better choice. What do you think?

fbourgey:
@steppi can you chime in here?

steppi:
I think it's fine to do this on a temporary basis. Longer term, we should move to only using fixed width types, using templates in the kernels for integer args to support "generic integer arguments", and to have `int32` and `int64` versions of the loops whenever there are integer args. I'm OK with merging this though.

dschmitz89:
> I think it's fine to do this on a temporary basis. Longer term, we should move to only using fixed width types, using templates in the kernels for integer args to support "generic integer arguments", and to have `int32` and `int64` versions of the loops whenever there are integer args. I'm OK with merging this though.

Would you have time to open a PR that show cases the template approach for the Jacobi polynomials? If that is the preferred route we should use it now already if it is not too complicated.

dschmitz89:
> > I think it's fine to do this on a temporary basis. Longer term, we should move to only using fixed width types, using templates in the kernels for integer args to support "generic integer arguments", and to have `int32` and `int64` versions of the loops whenever there are integer args. I'm OK with merging this though.
> 
> 

@steppi Would you have time to open a PR that show cases the template approach for the Jacobi polynomials? If that is the preferred route we should use it now already if it is not too complicated.

PR #221, opened 2026-07-21
BUG: sph_bessel_j: implement reflection for real (not just complex)

lucascolley:
Co-authored-by: Gert-Ludwig Ingold <gert.ingold@physik.uni-augsburg.de>

towards scipy/scipy#25668, gh-222. Companion PR: scipy/scipy#25675

```
In [1]: from timeit import timeit
   ...: import scipy
   ...: print(scipy.__version__)
   ...: from scipy.special import spherical_jn
   ...:
   ...: t_real = timeit("spherical_jn(10, 5.)", globals=globals())
   ...: t_cmplx = timeit("spherical_jn(10, 5.+0j)", globals=globals())
   ...: print(f"float:   {t_real:7.4f}")
   ...: print(f"complex: {t_cmplx:7.4f}")
1.19.0.dev0+git20260721.314809d
float:    0.9746
complex:  0.9945
```

Related bug report to `mpmath`: https://github.com/mpmath/mpmath/issues/1142


lucascolley:
@steppi if this looks right, @gertingold is happy to look into the other functions in this file

dschmitz89:
Thanks @lucascolley , the change looks good but let's add a test for it.  https://github.com/mpmath/mpmath/issues/319 contains a snippet to compute the function you need I think.

lucascolley:
> Thanks @lucascolley , the change looks good but let's add a test for it. [mpmath/mpmath#319](https://github.com/mpmath/mpmath/issues/319) contains a snippet to compute the function you need I think.

how/where do I add a test?

dschmitz89:
> > Thanks @lucascolley , the change looks good but let's add a test for it. [mpmath/mpmath#319](https://github.com/mpmath/mpmath/issues/319) contains a snippet to compute the function you need I think.
> 
> how/where do I add a test?

The correct folder is tests/xsf_tests. You can follow the pattern from my PR:  https://github.com/scipy/xsf/pull/218. That's actually about the same function partially

lucascolley:
thanks @dschmitz89, gave that a go. Looks like there are failures, not sure if the error is in the implementation or the tests

pixi.toml
```diff
@@ -47,6 +47,7 @@ xsref = { git = "https://github.com/scipy/xsref", rev = "1a5bb3d39efe4e9f26dc4da
 [feature.build.dependencies]
 cmake = "*"
 make = "*"
+ninja = "*"
 
 [feature.build.target.unix.dependencies]
 cxx-compiler = "*"
```
dschmitz89: Are these intentional?

dschmitz89 (commented):
I see the following error in the tests:

/home/runner/work/xsf/xsf/tests/xsf_tests/test_bessel.cpp:24: FAILED:
  REQUIRE( rel_err_spherical_j <= rtol )
with expansion:
  2.00000000000000355 <= 0.00000000000001
with messages:
  n := 10
  z := (-5,0)
  result_spherical_j := (0.000407344,0)
  ref_spherical_j := (-0.000407344,0)
  rel_err_spherical_j := 2.00000000000000355
  rtol := 0.00000000000001


So it looks like exactly the sign of one of them is wrong. Might also be the mpmath snippet. Could you confirm which is right using some reference? I guess your change is correct here.

pixi.toml
lucascolley: I got an error without having ninja on my machine, yeah. Does `pixi run build` work for you in an env without `ninja`?

lucascolley:
I figured out my mistake in generating the reference values, should be fixed now.

tests/xsf_tests/test_bessel.cpp
```diff
@@ -0,0 +1,43 @@
...
+    using test_case = std::tuple<long, std::complex<double>, std::complex<double>, double>;
+
+    // Reference values computed with mpmath.
+    auto [n, z, ref_spherical_j, rtol] = GENERATE(
+        // Even n => j_n(-z) = j_n(z)
+        test_case{10, {5., 0}, {0.00040734424424946041, 0}, 1e-14},
+        test_case{10, {-5., 0}, {0.00040734424424946041, 0}, 1e-14},
+        // Odd n => j_n(-z) = -j_n(z)
+        test_case{7, {5., 0}, {0.017902778177989527, 0}, 1e-14},
+        test_case{7, {-5., 0}, {-0.017902778177989527, 0}, 1e-14}
```
lucascolley: maybe we ought to link to https://dlmf.nist.gov/10.47#E14 here too

tests/xsf_tests/test_bessel.cpp
```diff
@@ -0,0 +1,43 @@
+#include "../testing_utils.h"
+#include <complex>
+#include <tuple>
+#include <xsf/bessel.h>
+#include <xsf/sph_bessel.h>
+
+TEST_CASE("spherical_j reflection complex", "[bessel][xsf_tests]") {
+    using test_case = std::tuple<long, std::complex<double>, std::complex<double>, double>;
+
+    // Reference values computed with mpmath.
```
lucascolley: I used the function defined in https://github.com/mpmath/mpmath/issues/319 however I believe it gets the reflection _wrong_ for the cases we test here. So I adjusted per the odd/even reflection we are fixing.

gertingold:
The reflection test in its present form looks fine to me. For even order, the spherical Bessel function of the first kind should even, for odd order it should be odd.

gertingold:
Just an observation: `mpmath` yields for even order an odd `spherical_jn`
```
>>> mp.mp.dps = 40
>>> n = 10
>>> z = 5
>>> mp.spherical_jn(n, z)
mpf('0.0004073442442494604294289103292570772538251032')
>>> mp.spherical_jn(n, -z)
mpc(real='-0.0004073442442494604294289103292570772538251032', imag='0.0')
```
which contradicts https://dlmf.nist.gov/10.47#E14. On the other hand, slightly off the real negative axis, `spherical_jn` behaves like a symmetric function as expected:
```
>>> mp.spherical_jn(n, -z+0.001j)
mpc(real='0.000407343696152771891947135307164805957966911', imag='-0.0000007219421519647832625003581949896496767654852')
>>> mp.spherical_jn(n, -z-0.001j)
mpc(real='0.000407343696152771891947135307164805957966911', imag='0.0000007219421519647832625003581949896496767654852')
>>> mp.spherical_jn(n, z-0.001j)
mpc(real='0.000407343696152771891947135307164805957966911', imag='-0.0000007219421519647832625003581949896496767654852')
>>> mp.spherical_jn(n, z+0.001j)
mpc(real='0.000407343696152771891947135307164805957966911', imag='0.0000007219421519647832625003581949896496767654852')
```

gertingold:
Should a test also be added for `sph_bessel_j_jac`, i.e. for the derivative of the spherical Bessel function of the first kind? Strictly speaking, this is not really necessary because the derivative is calculated from a recursion-type formula so that no reflection propery is actually used. On the other hand, it might be nice to be complete and verify the symmetry also for the derivative.

lucascolley:
Thanks, I opened https://github.com/mpmath/mpmath/issues/1142.

lucascolley:
@WarrenWeckesser any opinion on tests for this?

pixi.toml
dschmitz89: Yeah it works without locally and in CI. I did not update my pixi packages since the latest pixi changes though.

lucascolley: Ah I was testing out using ninja at some point in a different branch, probably if I wipe my build dir it would be okay without

dschmitz89: Yup, guess so. Revert this change and we can merge, if CI comes back green.

lucascolley: > Yup, guess so. Revert this change and we can merge, if CI comes back green.

Let's also address Gert's comment about testing `_jac` too: https://github.com/scipy/xsf/pull/221#issuecomment-5048675313

gertingold:
Today, I spent more time on analyzing the code. The results are as follows:

- `spherical_jn`
  - The PR already contains a solution to speed up the execution for real values.
  - The definition of `spherical_yn` might contain an even better solution but I did not have time yet to do timings. Probably, the difference is not too big anyway.
- `spherical_yn`
  - Here, the decorator for the reflection is not needed because the reflection was already handled in `xsf` before. The decorator can thus be removed. The real case is almost a factor of 2 faster than the complex case. That is why I expect that there is still some room of improvement for `spherical_jn`.
- `spherical_in`
  - Here, the situation resembles very much `spherical_jn`. It can thus be resolved correspondingly.
- `spherical_kn`
  - Since the modified spherical Bessel function of the second kind obeys a somewhat more complicated reflection formula, presently the decorator adds a vanishing imaginary part to the argument `z`. Of course, this could be done without the decorator, thereby speeding up execution time by a factor of 20 or so.

Before amending the PR, I will need to test my code, hopefully tomorrow.

lucascolley:
great — no rush!

dschmitz89:
I recommend to add one PR per function as small PRs are easier to review. So let's focus here on spherical _jn and it's derivatives. The follow up PRs for the other functions can take up the patterns from this PR then.

gertingold:
I have prepared a test for `sph_bessel_j_jac` both for real and complex arguments. This test is not expected to fail if `sph_bessel_j` works as intended. Therefore, I am wondering whether it makes sense to add it to `xsf` (it cannot really hurt) or whether this kind of test should be the reponsibility of tests in `scipy.special`. Essentially, I am uncertain about the status of `xsf`. Should it be considered as potentially stand alone or should it be always considered to be part of `scipy`?

Another small point: It might make sense to rename the new `test_bessel.py` into `test_sph_bessel.py` to make its purpose more obvious. After all, it tests the functions defined in `sph_bessel.h`.

And finally, if we decide to add the tests for `sph_bessel_j_jac`, what would be the correct way to add a commit to this PR?

lucascolley:
> I have prepared a test for `sph_bessel_j_jac` both for real and complex arguments. This test is not expected to fail if `sph_bessel_j` works as intended. Therefore, I am wondering whether it makes sense to add it to `xsf` (it cannot really hurt) or whether this kind of test should be the reponsibility of tests in `scipy.special`. Essentially, I am uncertain about the status of `xsf`. Should it be considered as potentially stand alone or should it be always considered to be part of `scipy`?

we can add it here 👍 `xsf` is intended to be usable by other libraries than SciPy.

> Another small point: It might make sense to rename the new `test_bessel.py` into `test_sph_bessel.py` to make its purpose more obvious. After all, it tests the functions defined in `sph_bessel.h`.

yes, good point

> And finally, if we decide to add the tests for sph_bessel_j_jac, what would be the correct way to add a commit to this PR?

You can submit a PR which targets my branch https://github.com/lucascolley/xsf/tree/bessel_j_reflect, then I can merge that and it will show up here.

dschmitz89:
One thing I want to note out of curiosity: over the last weeks we made a lot of improvements for our Bessel functions. While they cover mostly extreme inputs, I still had not expected that as Bessel functions are so common and algorithms for their computation have been around for a long time.

gertingold:
Just to give a little bit of background about how the present issue came up: In one of our projects, we had our own (pure Python) implementation for spherical Bessel functions employed to evaluate Mie scattering amplitudes. Comparing with the implementation in `scipy.special`, we noted that our code executes faster. This unexpected result led us to analyze the situation in more detail and we found a surprising difference in run time between real and complex arguments. In trying to understand the reason, I somehow got lost on the way from `scipy.special` to `xsf`. I felt much more at home with the old Fortran code than with C++. Discussing our findings with @lucascolley, he quickly came to the correct conclusion that the decorators taking care of the reflection property are the culprit.

lucascolley:
I believe this is ready from my side

dschmitz89 (approved):
Thanks @gertingold @lucascolley .

PR #210, opened 2026-07-04
ENH: Add `gen_harmonic` to XSF

khushi-411:
#### Reference issue

Follows: #170

#### What does this implement/fix?

- Port [scipy/special/gen_harmonic.h](https://github.com/scipy/scipy/blob/main/scipy/special/gen_harmonic.h) to XSF

#### AI Generation Disclosure

No AI tools used

PR #223, opened 2026-07-23
special: refactor common utilities into `config.h`

khushi-411:
#### Reference issue

Follow up of: https://github.com/scipy/xsf/pull/210#pullrequestreview-4651417395 and https://github.com/scipy/xsf/pull/210#discussion_r3534859084

- refactor common utilities into config.h

#### AI Generation Disclosure

No AI tools used.


lucascolley:
ping @j-bowhay

include/xsf/config.h
```diff
@@ -242,6 +242,12 @@ using pair = cuda::std::pair<T1, T2>;
 template <typename... Types>
 using tuple = cuda::std::tuple<Types...>;
 
+template <typename T1, typename T2>
+using is_same = cuda::std::is_same<T1, T2>;
+
+template <typename T1, typename T2>
+inline constexpr bool is_same_v = cuda::std::is_same_v<T1, T2>;
```
j-bowhay, line 249: How come these are needed?

khushi-411: @dschmitz89 mentioned in https://github.com/scipy/xsf/pull/210#discussion_r3534859084 that `is_same` won't work on GPU without defining it in `config.h`.

> From what I inderstand, is_same will not work directly on GPUs without adding it in config.h. For a first pass here, it should be fine to ignore the XSF_HOST_DEVICE macro as this is anyway a rather niche function that is not used yet on GPU.

Should I also add XSF_HOST_DEVICE to gen_harmonic? Since CuPy doesn't support it yet, I left it out for now. Otherwise, I can remove these aliases if you would prefer.

dschmitz89: My main problem with all this is that I cannot test the GPU compat myself due to the lack of the hardware and neither do we have GPU test coverage of these in CI. So whatever we do, any reviewer would have to do a lot of manual work or ask another reviewer.

How could we improve this for the Jacobi functions? Dies it make sense to add GPU tests of stuff not yet available in CuPy?

khushi-411: @dschmitz89, your point makes sense. I don't think it would be very useful to add GPU macros for functionality that isn't available in CuPy yet, since those code paths can't actually run.

However, adding GPU tests for the functionality that is supported in CuPy, and having CI run those tests, would be helpful.

steppi: Thanks everyone. Here's my understanding of the matter. The key things to keep in mind here are:

1. If a function `f` is in a header together with other functions that have GPU support set up, then `f` needs to have GPU support too. (Also if `f`'s header is included in a header that contains functions with GPU support).
2. If the function `f` really seems like it should go into a header with other functions which have GPU support, based on being in the same conceptual group, but it would be hard to add GPU support, then put `f` in a header with the same name under the cpu namespace (see `xsf/stats.h` and `xsf/cpu/stats.h`.
3. There's no need to wait for something to be in CuPy to add GPU support in `xsf`. CuPy doesn't work like SciPy where the ufuncs are built into the package. Anyone can use `xsf`'s kernels in CuPy by defining their own `ElementwiseKernel` or CuPy `ufunc`.
4. If you don't have access to a GPU for development purposes, don't worry about adding GPU support, but also keep in mind 1. and 2. above.

Since `gen_harmonic` isn't contained in a header with other functions that have GPU support, there's no need to add the macros here, and no need to use the cpu namespace. It can be useful to add `XSF_HOST_DEVICE` and any necessary additions to `config.h` for functions not available in CuPy though, it's just not a requirement for `xsf` contributors since not everyone has a compatible GPU.

dschmitz89: @steppi Thanks for the explanation. So my understanding is that the changes here do not need to be tested and do not cause any harm. In that light, good to merge?

khushi-411:
Thanks so much for the detailed explanation, @steppi, and thank you @dschmitz89 and @j-bowhay for the insightful questions and for reviewing the PR!

I've updated the PR to add `XSF_HOST_DEVICE` to `gen_harmonic`, as @steppi pointed out that it could be useful. Everything else has been addressed. Thanks!

include/xsf/config.h
steppi: > @steppi Thanks for the explanation. So my understanding is that the changes here do not need to be tested and do not cause any harm. In that light, good to merge?

Correct, these changes do no harm, and this should be good to merge. If it later turns out that this doesn't quite work in a CuPy `ElementwiseKernel` yet for one reason or another, that can always be fixed, but I think it's useful to make a best effort attempt to get a new function working in principle.

dschmitz89 (approved):

Issue #222, opened 2026-07-23
BUG: implement reflection for spherical bessel functions on the real negative axis


lucascolley:
Following https://github.com/scipy/xsf/pull/221, we should implement equivalent fixes for the `y`, `i`, and `k` functions too. cc @gertingold

gertingold:
I will try to find some time later today to work on it.

PR #225, opened 2026-07-26
TST: sph_bessel_y: added tests for reflection property

gertingold:
#### Reference issue
gh-222

This PR adds tests for the reflection property of the spherical Bessel functions of the second kind as was done before in #221 for the spherical Bessel functions of the first kind.

#### Additional information
In #221, the reflection in case of negative arguments `z` was added for the spherical Bessel functions of the first kind. For the Bessel functions of the second kind, the required code was already present, so that there is no immediate need to apply changes to `xsf/sph_bessel.h`.

However, in the variant for real argument `z`, the reflection is implemented with a code differing from the variant for complex argument and differing from the code used in `sph_bessel_j`.

The variant for real argument uses
https://github.com/scipy/xsf/blob/db2fdd0277b9cea80bf66781235400f51a6b5dd3/include/xsf/sph_bessel.h#L174-L176
while for complex arguments
https://github.com/scipy/xsf/blob/db2fdd0277b9cea80bf66781235400f51a6b5dd3/include/xsf/sph_bessel.h#L237-L240
is used. The corresponding variant for `sph_bessel_j` with real argument reads
https://github.com/scipy/xsf/blob/db2fdd0277b9cea80bf66781235400f51a6b5dd3/include/xsf/sph_bessel.h#L51-L54

I propose to use the latter code (with the function call adapted accordingly) in `sph_bessel_y` for real argument as well. Any thoughts on this? I will amend the PR accordingly, if the change shall be made.

Since the reflection property is already ensured for `sph_bessel_y`, the corresponding decorator in `scipy/special/_spherical_bessel.py` can be removed. This will be done in a separate PR once the present PR has been handled.

The changes for the modified spherical Bessel functions mentioned in gh-222 will be the subject of separate PRs.

#### AI Generation Disclosure
No AI tools used

### DIFF
--- a/include/xsf/sph_bessel.h
+++ b/include/xsf/sph_bessel.h
@@ -172,7 +172,8 @@ T sph_bessel_y(long n, T x) {
     }
 
     if (x < 0) {
-        return std::pow(-1, n + 1) * sph_bessel_y(n, -x);
+        T r = sph_bessel_y(n, -x);
+        return (n % 2 == 1) ? r : -r;
     }
 
     if (x == std::numeric_limits<T>::infinity() || x == -std::numeric_limits<T>::infinity()) {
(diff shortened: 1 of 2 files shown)


lucascolley:
> I propose to use the latter code (with the function call adapted accordingly) in `sph_bessel_y` for real argument as well. Any thoughts on this? I will amend the PR accordingly, if the change shall be made.

yes, agreed, the latter is arguably more readable in the context of reflection, even if it is further away from the formula found in DLMF.

dschmitz89 (approved):

Issue #228, opened 2026-07-29
CI/TST: very slow tests


lucascolley:
[a recent log](https://github.com/scipy/xsf/actions/runs/30428185957/job/90499182380) shows a few tests taking an exceedingly long amount of time on windows.

- [x] https://github.com/scipy/xsf/blob/main/tests/xsf_tests/test_hypergeometric.cpp:

```
379/379 Test #298: hyp2f1(-1, b, c, z) / gh-4446 ...................................   Passed   72.61 sec
```

Ping @fbourgey, looks like you touched that one last. I guess we should check that it isn't just due to it being the final test in the run?

- [ ] https://github.com/scipy/xsf/blob/main/tests/scipy_special_tests/test_cyl_bessel_i.cpp:

```
 29/379 Test  #30: cyl_bessel_i dd->d scipy_special_tests ..........................   Passed   49.92 sec
```

```
 54/379 Test  #29: cyl_bessel_i dD->D scipy_special_tests ..........................   Passed   52.53 sec
```

This is now the primary bottleneck for CI completion, so it would make an impact to fix these.

lucascolley:
note that these tests are slow on my osx-arm64 machine too, Windows just seems to exacerbate the problem

lucascolley:
For `cyl_bessel_i`, claude reckons it has an unusually large amount of entries in the xsref tables. If I guess correctly, that is due to https://github.com/scipy/scipy/commit/951961a73244175f8e4d3cbedbb0977fd90251b3.

lucascolley:
see also

```
  1/387 Test   #2: airy d->dddd scipy_special_tests ................................   Passed    6.58 sec
...
 12/387 Test   #1: airy D->DDDD scipy_special_tests ................................   Passed    7.08 sec
```

PR #230, opened 2026-07-30
TST: speed up hyp2f1 test

lucascolley:
Towards gh-228, patch is claude's suggestion. WDYT @fbourgey ?


fbourgey:
Thanks @lucascolley, this LGTM. Windows test runtimes appear to be less than 1s vs. 72s seconds before.

fbourgey (approved):

PR #231, opened 2026-07-31
BUG: fix y1 for large inputs

FireflySentinel:
#### Reference issue

Follow-up to j0 (scipy#131), y0 (scipy#140) and j1 (scipy#143). `y1` is the last one in that group still forming the phase as `x - 3pi/4`.

`xn = x - THPIO4` rounds to the nearest double, so it costs about `ulp(x)/2` of
absolute phase, which `sin`/`cos` turn into relative error.

I implemented the same rewrite as y0 (scipy#140), so the trig is evaluated on the unmodified `x`. I also adds the
missing `y1` case to `test_bessel_functions_large_inputs.cpp`.

#### Additional information

Relative error vs mpmath at 1000 dps:

| x | before | after |
|---|---|---|
| 1e4 | 1.1e-13 | 9.0e-18 |
| 1e8 | 1.4e-08 | 1.1e-16 |
| 1e20 | 6.7 | 1.2e-16 |

`j0`, `y0` and `j1` are at ~1e-16 on the same points, so this is `y1` only.
Additionally, `yv`/`yn` inherit it through the recurrence. For example, `yv(200, 1e7)` is off by 4.1e-10 before this change.

`pixi run tests` passes and the new test fails on main.

#### AI Generation Disclosure

I used Claude Code to draft the fix and test.

### DIFF
--- a/include/xsf/cephes/j1.h
+++ b/include/xsf/cephes/j1.h
@@ -199,9 +199,13 @@ namespace cephes {
         z = w * w;
         p = polevl(z, detail::j1_PP, 6) / polevl(z, detail::j1_PQ, 6);
         q = polevl(z, detail::j1_QP, 7) / p1evl(z, detail::j1_QQ, 7);
-        xn = x - detail::THPIO4;
-        p = p * std::sin(xn) + w * q * std::cos(xn);
-        return (p * detail::SQRT2OPI / std::sqrt(x));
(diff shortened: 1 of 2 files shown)


dschmitz89:
Thanks @FireflySentinel !

dschmitz89 (approved):

PR #234, opened 2026-08-03
ENH: add relection formula to sph_bessel_k

### LINKED ISSUE #222: BUG: implement reflection for spherical bessel functions on the real negative axis

j-bowhay:
#### Reference issue

closes #222 (I think, let me know if there is more to do)

Adds relection formula to `sph_bessel_k` which is given by

$$k_n(-z)= -\frac{\pi}{2}\left(i_n^{(n)}(z) + i_n^{(2)}(z)\right)$$

except that we don't have an $i_n^{(2)}$ so I use the property that

$$k_n(z) = (-1)^n\frac{\pi}{2}\left(i_n^{(1)}(z) - i_n^{(2)}(z)\right)$$

which gives

$$k_n(-z)= -\frac{\pi}{2}\left(2i_n^{(1)}(z) + (-1)^n\frac{2}{\pi}k_n(z)\right)$$
#### AI Generation Disclosure

I used Claude to double check the derivation.

### DIFF
--- a/tests/xsf_tests/test_sph_bessel.cpp
+++ b/tests/xsf_tests/test_sph_bessel.cpp
@@ -246,41 +246,6 @@ TEST_CASE("spherical_i tiny inputs", "[spherical_bessel][xsf_tests]") {
     REQUIRE(rel_err <= rtol);
 }
 
-TEST_CASE("spherical_k reflection complex", "[spherical_bessel][xsf_tests]") {
-    using test_case = std::tuple<long, std::complex<double>, std::complex<double>, double>;
-
-    // Reference values computed with mpmath.
-    auto [n, z, ref_spherical_k, rtol] = GENERATE(
-        test_case{10, {5., 0}, {11.162178171949055, 0}, 1e-14},
-        test_case{10, {-5., 0}, {-11.165977657150502, 0}, 1e-14},
-        test_case{7, {5., 0}, {0.22221361309239493, 0}, 1e-14},
-        test_case{7, {-5., 0}, {-0.023872188945034639, 0}, 1e-14}
-    );
-
-    std::complex<double> result_spherical_k = xsf::sph_bessel_k(n, z);
-    double rel_err_spherical_k = xsf::extended_relative_error(result_spherical_k, ref_spherical_k);
-
-    CAPTURE(n, z, result_spherical_k, ref_spherical_k, rel_err_spherical_k, rtol);
-    REQUIRE(rel_err_spherical_k <= rtol);
-}
-
-TEST_CASE("spherical_k_jac reflection derivative complex", "[spherical_bessel][xsf_tests]") {
-    using test_case = std::tuple<long, std::complex<double>, std::complex<double>, double>;
-
-    // Reference values computed with mpmath.
-    auto [n, z, ref_spherical_k_jac, rtol] = GENERATE(
-        test_case{10, {5., 0}, {-27.299149693641313, 0}, 1e-14},
-        test_case{10, {-5., 0}, {-27.290758018530437, 0}, 1e-14},
-        test_case{7, {5., 0}, {-0.43011979527682201, 0}, 1e-14}, test_case{7, {-5., 0}, {0.84209146503609691, 0}, 1e-14}
-    );
-
-    std::complex<double> result_spherical_k_jac = xsf::sph_bessel_k_jac(n, z);
-    double rel_err_spherical_k_jac = xsf::extended_relative_error(result_spherical_k_jac, ref_spherical_k_jac);
-
-    CAPTURE(n, z, result_spherical_k_jac, ref_spherical_k_jac, rel_err_spherical_k_jac, rtol);
-    REQUIRE(rel_err_spherical_k_jac <= rtol);
-}
-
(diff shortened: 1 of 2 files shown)


j-bowhay:
cc @lucascolley @gertingold

tests/xsf_tests/test_sph_bessel.cpp
```diff
@@ -245,3 +245,17 @@ TEST_CASE("spherical_i tiny inputs", "[spherical_bessel][xsf_tests]") {
     CAPTURE(n, z, result, ref, rel_err, rtol);
     REQUIRE(rel_err <= rtol);
 }
+
+TEST_CASE("spherical_k real reflection", "[spherical_bessel][xsf_tests]") {
```
lucascolley, line 249: can we test complex and `_jac` like the others?

gertingold: As unlike the other (modified) spherical Bessel functions, $k_n$ does not have an even or odd reflection symmetry, the approach taken in `scipy/special/_spherical_bessel.py` for arguments on the real negative axis is to add a vanishing imaginary part and make use of the code for complex arguments. I was trying to move this approach to `xsf`, but not being a C++ expert, I did not succeed so far. Somehow, calling the complex version from the real version did not work because the correct version of the function was not found. As far as the tests are concerned, I have already prepared the tests for k with real and imaginary arguments based on the results from `mpmath`. I will push the work done so far to my fork tonight, so in case somebody has time to take a look at it and can give advise concerning the C++ code, it would be great. Writing the test for `_jac` should be straightforward and I can certainly work on this.

A remark concerning the comment by @j-bowhay: The problem is that `spherical_in` yields $i_n^{(1)}$, but as far as I know, there is no way to obtain $i_n^{(2)}$ with Scipy. Of course, $i_n^{(2)}$ can be obtained from $i_n^{(1)}$ and $k_n$, but then we are turning in circles. But may be I am missing something.

gertingold: Sorry, it is too hot in my office and I got confused by the superscript `(n)` in the last equation of the comment by @j-bowhay. Of course, one can eliminate $i_n^{(2)}$. However, I am wondering whether making two calls to modified spherical Bessel functions instead of a single one, even for complex argument, makes sense. After all, the runtime for the (modified) spherical Bessel functions of the first kind does not depend much on whether the argument is real or complex.

j-bowhay: Indeed there was a typo!

j-bowhay: I don't have a strong opinion on which approach to take

j-bowhay:
Hmm seems like there are some platform dependent failures to look into

j-bowhay:
Although funnily enough the test failures are for the complex implementation which is not touched in this PR...

gertingold:
Strange, it looks like it is only a sign issue, but the sign should not depend on the platform.

j-bowhay:
> Strange, it looks like it is only a sign issue, but the sign should not depend on the platform.

Would you mind if I raised a separate issue about the complex inputs with your tests? I think the changes added here are correct so shouldn't be blocked by this issue since it doesn't originate from this PR.

gertingold:
Go ahead. I agree that this problem is not related to your changes.

I just checked the sign in the new tests against the sign which one obtains from the present `scipy.spherical_kn` and everything fits, even for the derivative.

gertingold:
A remark concerning the failing tests: 
- `macos`: Here, only one test (n=7, z=-5) fails because the threshold of the relative error of 1e-14 is somewhat exceeded. The other test (n=10, z=-5) as well as the tests for the derivative for both cases pass fine. 
- `arm`: Here, the sign for real negative z is incorrect so that all test cases with negative argument including the derivative fail.

tests/xsf_tests/test_sph_bessel.cpp
```diff
@@ -245,3 +245,70 @@ TEST_CASE("spherical_i tiny inputs", "[spherical_bessel][xsf_tests]") {
...
+        test_case{10, {-5., 0}, {-27.290758018530437, 0}, 1e-14},
+        test_case{7, {5., 0}, {-0.43011979527682201, 0}, 1e-14}, test_case{7, {-5., 0}, {0.84209146503609691, 0}, 1e-14}
+    );
+
+    std::complex<double> result_spherical_k_jac = xsf::sph_bessel_k_jac(n, z);
+    double rel_err_spherical_k_jac = xsf::extended_relative_error(result_spherical_k_jac, ref_spherical_k_jac);
+
+    CAPTURE(n, z, result_spherical_k_jac, ref_spherical_k_jac, rel_err_spherical_k_jac, rtol);
+    REQUIRE(rel_err_spherical_k_jac <= rtol);
+}
```
j-bowhay:
```suggestion
```

j-bowhay:
OK I removed the complex tests and opened https://github.com/scipy/xsf/issues/237. Hopefully this will now be ready to go

lucascolley (approved):
Thanks both, hopefully we can figure out what's going on with ARM in a follow-up

Issue #232, opened 2026-08-02
BUG: kelvin: the shared |x| < 10 branch point costs ber/bei/berp/beip ~6 digits, and ker/kei return NaN at the domain edges


EazyHood:
All numbers below were measured on an installed **scipy 1.17.1** against **mpmath at 50 digits**, not emulated. Everything is reproducible with the snippets included.

`include/xsf/kelvin.h`, `detail::klvna`, line 48:

```cpp
if (fabs(x) < 10.0) {
```

This single branch point selects between the ascending series and the asymptotic expansion **for all eight outputs at once** — `ber`, `bei`, `ker`, `kei` and their derivatives. The eight do not want the same crossover, and fixing it at 10 is right for one half of them and costly for the other.

## 1. Six digits lost at the seam

`ber` is accurate to ~3 ulps immediately below the branch point and loses six digits immediately above it:

```python
>>> import scipy.special as sp, mpmath as mp
>>> mp.mp.dps = 50
>>> sp.ber(9.999999999999998)
138.84046594163263        # mpmath: 138.84046594163255627   rel 5.29e-16
>>> sp.ber(10.0)
138.84046601322478        # mpmath: 138.84046594163264721   rel 5.16e-10
```

Those two arguments are adjacent doubles, 1.8e-15 apart. `ber` moves 9.09e-14 between them. The returned values move **7.16e-08** — a factor of **787,000**.

Same seam, same size, at x = 10.0:

| | scipy | mpmath (50 dps) | rel. error |
|---|---|---|---|
| `ber` | 138.84046601322478 | 138.84046594163264721 | 5.16e-10 |
| `bei` | 56.37045856165942 | 56.370458553906638233 | 1.38e-10 |
| `berp` | 51.19525834615495 | 51.195258393599611382 | 9.27e-10 |
| `beip` | 135.3093016566432 | 135.30930171596461358 | 4.38e-10 |

Forcing each branch separately shows the series is still the better choice well past 10 for these four:

| x | series | asymptotic |
|---|---|---|
| 12 | 6.7e-16 | 9.4e-12 |
| 14 | 5.8e-16 | 8.6e-13 |
| 16 | 6.7e-15 | 4.8e-14 |
| 18 | 6.2e-16 | 7.0e-15 |

The measured crossover — the first x where the asymptotic matches the series — is 20.0 for `ber`, 18.0 for `bei`, 19.0 for `berp` and `beip`. So for these four the branch point is roughly nine units too early.

[...]

PR #236, opened 2026-08-04
TST(kelvin): kelvin: add a Wronskian identity test

EazyHood:
Follows up on the Wronskian test offered in gh-232, which @steppi said he'd be happy to see.

## What it checks

The Kelvin functions satisfy

```
x * (Be*Kep - Bep*Ke) == -1
```

with `Be = ber + i*bei`, `Ke = ker + i*kei`, and `Bep`, `Kep` their derivatives — exactly the four values `xsf::kelvin()` returns. Its real part is `(ber*kerp - berp*ker) - (bei*keip - beip*kei)` and its imaginary part is `(ber*keip - berp*kei) + (bei*kerp - beip*ker)`, so a single check exercises all eight functions.

I derived it by searching bilinear combinations against `mpmath` rather than quoting it: over `x` in `[0.05, 25]` at 50 digits the largest deviation is **1.4e-51**, which is mpmath's own rounding.

**It needs no reference table** — the functions are checked against each other. That is why it does not depend on how the `|x| < 10` branch point question in gh-232 is resolved.

## What it measures on the current implementation

Sweeping 52 points against `scipy` 1.17.1:

| x | \|residual\| |
|---|---|
| 0.001 | 0 |
| 0.5 | 0 |
| 5.0 | 7.1e-14 |
| **9.9** | **8.2e-11** |
| 10.0 | 7.4e-13 |
| 20.0 | 2.2e-16 |
| 50.0 | 0 |

The worst point in the whole sweep sits **just below the `|x| < 10` branch point**, and crossing it improves by ~111x. That is the claim gh-232 makes, arrived at without mpmath and without any reference values of mine — it is the library disagreeing with itself.

## On the tolerance

`atol = 1e-9` passes today with about two orders of margin, so this lands as a non-regression test rather than a failing assertion. At `1e-12` it fails at `x = 9.9` on current `main`. I chose the looser bound deliberately so the test is mergeable independently of the branch-point work; if you would rather it fail loudly and block until that is fixed, say so and I will tighten it.

## What I have not done

**I have not compiled this.** Catch2 3, Arrow and Parquet are not available on the machine I measured from, so the numbers above are verified but the build is not. It goes in `tests/xsf_tests/` rather than `tests/scipy_special_tests/` because it needs no parquet tables, and `CMakeLists.txt` picks it up through the existing `file(GLOB TEST_SOURCES "*/test_*.cpp")` — no build-system change. If CI finds a compile error I will fix it promptly; the one thing I would guess at is whether `CAPTURE()` prints `std::complex<double>`, which is diagnostic only.

---

AI disclosure: prepared with AI assistance (Claude Opus 5 via Claude Code). The identity and every figure above were measured, and I am responsible for the change as submitted.

### DIFF
--- a/tests/xsf_tests/test_kelvin_wronskian.cpp
+++ b/tests/xsf_tests/test_kelvin_wronskian.cpp
@@ -1,0 +1,40 @@
+#include "../testing_utils.h"
+
+#include <xsf/kelvin.h>
+
+// Wronskian of the Kelvin equation, in the complex form the functions come in:
+//
+//     x * (Be*Kep - Bep*Ke) == -1
+//
+// with Be = ber + i*bei, Ke = ker + i*kei, and Bep, Kep their derivatives --
+// exactly what xsf::kelvin() returns. Its real part is
+// (ber*kerp - berp*ker) - (bei*keip - beip*kei) and its imaginary part is
+// (ber*keip - berp*kei) + (bei*kerp - beip*ker), so one check covers all eight
+// functions at once.
+//
+// Verified against mpmath at 50 digits for x in [0.05, 25]: the largest deviation
+// is 1.4e-51, i.e. mpmath's own rounding.
+//
+// This needs no reference table -- the eight functions are checked against each
+// other -- so it stays valid whatever is decided about the |x| < 10 branch point
+// discussed in gh-232.
+
+TEST_CASE("kelvin wronskian", "[ber][bei][ker][kei][kelvin][xsf_tests]") {
+    // Loose enough to pass today. The worst residual measured on the current
+    // implementation is 8.2e-11, at x = 9.9 -- just below the |x| < 10 branch
+    // point, and about 111x worse than the first point above it.
+    const double atol = 1e-9;
+
+    const auto x = GENERATE(
+        0.001, 0.01, 0.1, 0.5, 1.0, 2.0, 3.0, 5.0, 7.0, 9.0, 9.5, 9.9,
+        10.0, 10.1, 10.5, 11.0, 15.0, 20.0, 30.0, 50.0
+    );
+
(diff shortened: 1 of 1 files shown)


tests/xsf_tests/test_kelvin_wronskian.cpp
```diff
@@ -0,0 +1,39 @@
...
+//     x * (Be*Kep - Bep*Ke) == -1
+//
+// with Be = ber + i*bei, Ke = ker + i*kei, and Bep, Kep their derivatives --
+// exactly what xsf::kelvin() returns. Its real part is
+// (ber*kerp - berp*ker) - (bei*keip - beip*kei) and its imaginary part is
+// (ber*keip - berp*kei) + (bei*kerp - beip*ker), so one check covers all eight
+// functions at once.
+//
+// Verified against mpmath at 50 digits for x in [0.05, 25]: the largest deviation
+// is 1.4e-51, i.e. mpmath's own rounding.
```
j-bowhay: What does this comment refer to?

tests/xsf_tests/test_kelvin_wronskian.cpp
```diff
@@ -0,0 +1,39 @@
...
+//
+// This needs no reference table -- the eight functions are checked against each
+// other -- so it stays valid whatever is decided about the |x| < 10 branch point
+// discussed in gh-232.
```
j-bowhay:
```suggestion
```

EazyHood:
Both removed. The mpmath line described a separate check I ran while writing the test, not
anything the test does - the Wronskian is self-checking, so that number didn't belong in the
header.

The workflows on this PR are all sitting in action_required and have never run, so there is no
CI signal on it yet. Could someone approve the run when convenient?

PR #239, opened 2026-08-05
ENH: add kernel for the `trigamma` function

j-bowhay:
#### Reference issue

Towards https://github.com/scipy/scipy/issues/7410

I have translated the Julia implementation of `trigamma` https://github.com/JuliaMath/SpecialFunctions.jl/blob/c9ff36085c9008b9bdaa457e745f459a1827e52e/src/gamma.jl#L58-L62 into an `xsf` kernel.
#### AI Generation Disclosure

I used chatgpt to work out how to plumb in the different real and complex polynomial evaluation.


include/xsf/trigamma.h
```diff
@@ -0,0 +1,85 @@
...
+    T w = t * t;
+    psi += t + 0.5 * w;
+    double coeffs[] = {-7.092156862745098,   1.1666666666666667,   -0.2531135531135531,  0.07575757575757576,
+                       -0.03333333333333333, 0.023809523809523808, -0.03333333333333333, 0.16666666666666666};
+    using base_t = std::remove_cv_t<T>;
+    if constexpr (std::is_same_v<base_t, std::complex<double>> || std::is_same_v<base_t, std::complex<float>>) {
+        psi += t * w * xsf::cevalpoly(coeffs, 7, w);
+    } else {
+        psi += t * w * xsf::cephes::polevl(w, coeffs, 7);
+    }
```
j-bowhay: In a follow-up PR I'll clean this up but that would involve touching other parts of `xsf` which I didn't want to do here

dschmitz89:
From a first look this seems pretty solid already. Two general questions below.

How does this compare to the trigamma implementation in #123 ?

I am not sure how difficult it is to do in pure C++ but could you create a log spaced plot of the complex trigamma function using this implementation? I basically want to make sure that we are not missing anything obvious as I find that hard to evaluate purely with unit tests. As we are using the same algorithms as Julia we are likely safe but one never knows.

j-bowhay:
Hi @dschmitz89 sorry it took so long for me to get back to you...... Here are the plots.
The first is comparing the error on the positive real line against our current implemention:
image
So pretty comparable!

Next, points in the complex plane (the whole point of this pr):
image

Here is my script:
```
import ctypes
import subprocess
import tempfile
from pathlib import Path

import matplotlib.pyplot as plt
import numpy as np
from scipy.special import polygamma
from mpmath import mp
mp.dps = 50


class Complex128(ctypes.Structure):
    _fields_ = [("real", ctypes.c_double), ("imag", ctypes.c_double)]

def load_trigamma():
    repo = Path(__file__).resolve().parent
    source = r'''
#include "xsf/trigamma.h"

extern "C" {

struct Complex128 { double real; double imag; };

Complex128 local_trigamma(Complex128 zin) {
    std::complex<double> z(zin.real, zin.imag);
    std::complex<double> r = xsf::trigamma(z);
    return Complex128{r.real(), r.imag()};
}

}
'''
    build_dir = tempfile.TemporaryDirectory(prefix="xsf-25287-")
    source_path = Path(build_dir.name) / "trigamma.cpp"
    library_path = Path(build_dir.name) / "trigamma.so"
    source_path.write_text(source)
    subprocess.run(
        ["c++", "-std=c++17", "-shared", "-fPIC", f"-I{repo / 'include'}", source_path, "-o", library_path],
        check=True,
    )
    library = ctypes.CDLL(str(library_path))
    function = library.local_trigamma
    function.argtypes = [Complex128]
    function.restype = Complex128
    return build_dir, function


def main():
    x = np.logspace(0, np.log10(np.finfo(float).max), 10000)
    ref = np.array([float(mp.psi(1, z)) for z in x])

    build_dir, local_trigamma = load_trigamma()

    def trigamma(z):
        z = complex(z)
        r = local_trigamma(Complex128(z.real, z.imag))
        return complex(r.real, r.imag)

    try:
        local_values = np.array([trigamma(z) for z in x])
    finally:
        build_dir.cleanup()

    plt.hist(np.abs(local_values - ref), bins=20, log=True, label="new", alpha=0.5)
    plt.hist(np.abs(polygamma(1, x) - ref), bins=20, log=True, label="scipy", alpha=0.5)
    plt.title("Real line")
    plt.legend()
    plt.show()

    rng = np.random.default_rng(29839238579284759)
    n = 1000
    z = rng.uniform(-1e5, 1e5, size=n) + 1j * rng.uniform(-1e5, 1e5, size=n)
    ref = np.empty_like(z)
    res = np.empty_like(ref)
    for i, z in enumerate(z):
        print(i)
        ref[i] = complex(mp.psi(1, z))
        res[i] = trigamma(z)
    plt.hist(np.abs(res - ref), bins=20, log=True, label="new", alpha=0.5)
    plt.show()


if __name__ == "__main__":
    main()
```

j-bowhay:
Hmmm probably need to handle nan inputs explicitly

j-bowhay:
> How does this compare to the trigamma implementation in #123 ?

Oops I completely missed this. They use a lot of similar ideas, I guess I like this approach because it has already been battle tested in Julia for a few years

include/xsf/trigamma.h
```diff
@@ -0,0 +1,90 @@
...
+        }
+        z += n;
+    }
+
+    // https://dlmf.nist.gov/5.15.E8
+    T t = 1.0 / z;
+    T w = t * t;
+    psi += t + 0.5 * w;
+    // order of the Bernoulli numbers is reversed due to `evalpoly` convention
+    double coeffs[] = {-7.092156862745098,   1.1666666666666667,   -0.2531135531135531,  0.07575757575757576,
```
dschmitz89, line 73:
```suggestion
    constexpr double coeffs[] = {-7.092156862745098,   1.1666666666666667,   -0.2531135531135531,  0.07575757575757576,
```

include/xsf/trigamma.h
```diff
@@ -0,0 +1,90 @@
...
+                       -0.03333333333333333, 0.023809523809523808, -0.03333333333333333, 0.16666666666666666};
+    using base_t = std::remove_cv_t<T>;
+    psi += t * w * evalpoly(coeffs, 7, w);
```
dschmitz89, line 76: The coefficient vector has length 8. Is it 7 because of indexing starting at 0 (never used the `evalpoly` function)?

include/xsf/trigamma.h
```diff
@@ -0,0 +1,90 @@
...
+// digamma function, for complex argument, Computer Physics Communications, 1972
+// https://doi.org/10.1016/0010-4655(72)90012-4.
+template <typename T>
+XSF_HOST_DEVICE inline T trigamma(T z) {
+    double x = std::real(z);
+    if (x <= 0) {
+        // reflection formula http://dlmf.nist.gov/5.15.E6
+        T s = sinpi(z);
+        T c = M_PI / s;
+        return c * c - trigamma(T(1) - z);
```
dschmitz89, line 51: Might be a matter of taste but I always try to avoid recursion. Here, it would be simpler to define a function `trigamma_positive_real(x)` for the path below and call that. From what I know, recursion is also not ideal for GPUs (not that one recursion will likely have a large impact).

dschmitz89 (changes requested):
A few inline comment and one about tests. Could you add consistency tests against `polygamma` and one for `float` inputs?

include/xsf/trigamma.h
dschmitz89: Reading the code again, what I wrote is likely nonsense. Feel free to ignore.

lucascolley:
cc @LorenzoPeri17, how do you think this compares to your approach?

LorenzoPeri17:
Thanks for the cc! The approach is very clean and from a quick benchmark is comparable accuracy and slightly faster implementation than mine! I'd be happy to rescope #123 to just the higher-order polygammas (which as far as I can tell is missing from this one :)

PR #202, opened 2026-06-19
BLD/DEV/CI: add and use Pixi package definitions

lucascolley:
towards gh-24

The rough idea:
- the various `build` tasks in our workspace become proper Pixi package definitions which produce an `xsf` package which can be depended on in a Pixi workspace
- for variants of `xsf` for which we build the tests, we now install the tests in the package under `share/xsf/tests/` rather than leaving them in the build directory
- the top-level workspace definition then just needs to hold non-building tasks, and environment definitions which use these variants of `xsf`

PR #242, opened 2026-08-06
CI: run GPU job on cirun on-prem runner

aktech:
Switch the CuPy GPU job from the Cirrus GPU runner to a cirun on-prem runner. Replace cirruslabs/cache with actions/cache and enable setup-pixi's built-in cache.

#### Reference issue
Follow up from #25664

Switch the CuPy GPU job from the Cirrus GPU runner to a cirun on-prem runner. Replace cirruslabs/cache with actions/cache and enable setup-pixi's built-in cache.

#### Additional information
Runner is provisioned via the `cirun-scipy-onprem-gpu--<run_id>` label; actions/cache is pinned to the v6.1.0 commit and keeps the existing run_id key + restore-keys scheme.

#### AI Generation Disclosure
Claude Code (Fable 5) was used to draft the workflow changes. I take full responsibility of the changes.

### DIFF
--- a/.github/workflows/gpu_tests.yml
+++ b/.github/workflows/gpu_tests.yml
@@ -5,9 +5,6 @@ on: [ push, pull_request ]
 permissions:
    contents: read  # to fetch code (actions/checkout)
 
-env:
-  CCACHE_DIR: "${{ github.workspace }}/.ccache"
-
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
@@ -21,16 +18,6 @@ jobs:
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           submodules: recursive
-
-      - name: Setup compiler cache
-        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
-        with:
-          path: ${{ env.CCACHE_DIR }}
-          # Make primary key unique by using `run_id`, this ensures the cache
-          # is always saved at the end.
-          key: ${{ runner.os }}-gpu-ccache-${{ github.run_id }}
-          restore-keys: |
-            ${{ runner.os }}-gpu-ccache
 
       - name: run nvidia-smi
         run: nvidia-smi
--- a/.github/workflows/gpu_tests.yml
+++ b/.github/workflows/gpu_tests.yml
@@ -15,7 +15,7 @@ concurrency:
 jobs:
   cupy_tests:
     name: CuPy GPU
-    runs-on: ghcr.io/cirruslabs/ubuntu-runner-amd64-gpu:24.04
+    runs-on: "cirun-scipy-onprem-gpu--${{ github.run_id }}"
     steps:
       - name: Checkout xsf repo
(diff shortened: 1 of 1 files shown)


lucascolley:
is the wait expected?

aktech:
Nope, looks like the repository is not activated in cirun. I can toggle that.

lucascolley:
I approved CI

.github/workflows/gpu_tests.yml
```diff
@@ -15,15 +15,15 @@ concurrency:
...
       - name: Checkout xsf repo
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           submodules: recursive
 
       - name: Setup compiler cache
-        uses: cirruslabs/cache@v4  #caa3ad0624c6c2acd8ba50ad452d1f44bba078bb # v4
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
           path: ${{ env.CCACHE_DIR }}
```
lucascolley: log says:

```
[warning]Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
```

https://github.com/scipy/xsf/actions/runs/31204654367/job/93085536720?pr=242

lucascolley (changes requested):

lucascolley:
It looks like the GPU CI is completely dead on main now, stalling for 24 hours then being cancelled: https://github.com/scipy/xsf/actions/workflows/gpu_tests.yml.

Would be great to push this to completion when you can @aktech !

aktech:
Hey! Sure I'll find some time today to wrap this off

.github/workflows/gpu_tests.yml
aktech: It doesn't look like we are doing any C++ compilation in this workflow, which means this caching step is pointless, for pixi we already have a cache in the pixi action.

I am going to remove this step entirely, someone might have copied it from the main scipy repo accidentally. Let me know if you think otherwise.

lucascolley: sure go for it!

aktech: done

lucascolley:
> It doesn't look like we are doing any C++ compilation in this workflow, which means this caching step is pointless, for pixi we already have a cache in the pixi action.

Hmm, on second thought, we aren't doing any compilation on a warm cache, but I think we do need to recompile if a commit edits the tests. Do you agree?

lucascolley:
ah never mind, ignore me, this is just the pytest part

lucascolley (approved):
thanks Amit!

lucascolley:
hey @aktech, as you can see at https://github.com/scipy/xsf/actions/runs/32855746336/job/97831382948 the job is now dominated by cache restore time (given the updates in my PR gh-202). Is there any way to speed that up?

aktech:
Yes, correct. We can set this up with s3 compatible caching (something like https://docs.cirun.io/caching/s3-compatible#cloudflare-r2-recommended-for-on-prem), I'll create an example PR, you'd need to add some creds in repo secrets, I can share privately.

lucascolley:
great, thanks!

rgommers:
> you'd need to add some creds in repo secrets, I can share privately.

thanks! added to repository secrets for this repo now

PR #243, opened 2026-08-06
MAINT: improved polynomial evaluation interface

j-bowhay:

While working on #239 it became quite annoying that there are two completely different functions for evaluating a polynomial with either a real or complex input. This PR introduces a unified `xsf::evalpoly` which has a `double` and `std::complex<double>` overload.
#### AI Generation Disclosure

No ai

### DIFF
--- a/include/xsf/evalpoly.h
+++ b/include/xsf/evalpoly.h
@@ -19,11 +19,12 @@
 
 #pragma once
 
+#include "cephes/polevl.h"
 #include "config.h"
 
 namespace xsf {
 
-XSF_HOST_DEVICE inline std::complex<double> cevalpoly(const double *coeffs, int degree, std::complex<double> z) {
+XSF_HOST_DEVICE inline std::complex<double> evalpoly(const double *coeffs, int degree, std::complex<double> z) {
     /* Evaluate a polynomial with real coefficients at a complex point.
      *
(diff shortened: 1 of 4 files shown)


j-bowhay:
@fbourgey would you mind taking a look :)

include/xsf/evalpoly.h
```diff
@@ -19,11 +19,12 @@
 
 #pragma once
 
+#include "cephes/polevl.h"
 #include "config.h"
 
 namespace xsf {
 
-XSF_HOST_DEVICE inline std::complex<double> cevalpoly(const double *coeffs, int degree, std::complex<double> z) {
+XSF_HOST_DEVICE inline std::complex<double> evalpoly(const double *coeffs, int degree, std::complex<double> z) {
```
fbourgey, line 27: What about if degree=0? Should we add a guard

```c++
if (degree=0) {
   return coeffs[0];
}
```
as coeffs[1] might not exist. The same would apply for the float overload.

j-bowhay: Done - there is quite a bit else that could be done such as checking degree is non-negative and adding standalone tests but I would rather not grow the scope of this PR

fbourgey:
LGTM, thanks!

PR #244, opened 2026-08-06
DOC: add permalinks for scipy/special/_precompute/ scripts

fbourgey:
As I was working on removing some old scripts in scipy/special/_precompute, I thought it would be good to add permalinks to those scripts in XSF.

The same thing exists for cosine.h https://github.com/scipy/xsf/blob/main/include/xsf/cosine.h#L42

#### AI Generation Disclosure

Asked Codex to check what I did and if I missed any.

### DIFF
--- a/include/xsf/hyp2f1.h
+++ b/include/xsf/hyp2f1.h
@@ -56,7 +56,8 @@ namespace detail {
     /* The original implementation in SciPy from Zhang and Jin used 1500 for the
      * maximum number of series iterations in some cases and 500 in others.
      * Through the empirical results on the test cases in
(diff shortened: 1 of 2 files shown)


lucascolley (approved):
awesome, thanks Florian! I'll ignore the GHA blip

fbourgey:
@steppi is it then OK to remove some of the scripts from https://github.com/scipy/scipy/tree/main/scipy/special/_precompute? I was thinking of `lambertw.py`, `loggamma.py`, `wrightomega.py`, `zetac.py` to start with. Or are you thinking of moving those somewhere else?

PR #247, opened 2026-08-08
MAINT: make correct sign of coefficients responsibility of caller of mathieu sum_fourier_series

steppi:

This PR moves the sign correction logic out of `xsf::mathieu::sum_fourier_series`. As revealed here, https://github.com/scipy/scipy/pull/25128#pullrequestreview-4874095191, proper sign correction needs to be aware of `m` and `q`, since the easy to understand correction that is currently employed is not always well conditioned. I am instead adding the sign correction on the SciPy side in https://github.com/scipy/scipy/pull/25128.

#### AI Generation Disclosure

No AI


steppi:
I'm going to go ahead and merge this.

PR #253, opened 2026-08-19
ENH(kolmogorov): template `kolmogorov` for `float` and `double`

fbourgey:
Replace the separate float and double overloads with a constrained template for kolmogorov. This PR tries to replicate was is  done in https://github.com/scipy/xsf/blob/main/include/xsf/tools.h#L45. 

Note that https://github.com/scipy/xsf/blob/main/include/xsf/gamma.h#L11 appears to have no constraints.

#### AI Generation Disclosure

Asked Codex for feedback.

### DIFF
--- a/include/xsf/config.h
+++ b/include/xsf/config.h
@@ -339,3 +339,10 @@ using complex = complex_type_t<T>;
 } // namespace xsf
 
 #endif
+
+namespace xsf {
+
+template <typename T>
+inline constexpr bool is_float_or_double_v = std::is_same_v<T, float> || std::is_same_v<T, double>;
+
+} // namespace xsf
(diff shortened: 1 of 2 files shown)


fbourgey:
Another solution is to define in xsf/config.h

```c++
template <typename T>
using enable_if_float_or_double_t = typename std::enable_if<is_float_or_double_v<T>, int>::type;
```

and use 

```c++
template <typename T, enable_if_float_or_double_t<T> = 0>
    inline T kolmogorov(T x) {
        return static_cast<T>(cephes::kolmogorov(static_cast<double>(x)));
    }
```

@steppi, @j-bowhay @dschmitz89 any thoughts?

fbourgey:
@steppi can you please take a look?

PR #255, opened 2026-08-19
BUG(par_cyl): pbdv: fix for large arguments

verma-divyanshu-git:
## Summary

- use the accurate gamma and trigonometric helpers near integer orders
- use the connection formula where the forward recurrence loses accuracy
- avoid premature underflow when normalizing the backward recurrence
- add regression coverage for the reported large-argument failures

Addresses scipy/scipy#25814.

## Tests

- `pixi run tests`
- `pixi run format-dry-error`

## AI disclosure

I used AI assistance to understand the repository structure and generate an initial draft of the potential fix and regression tests.
The initial drafts of v small parts of `include/xsf/par_cyl.h` and `tests/scipy_special_tests/test_pbdv.cpp` were AI-assisted.
I reproduced the reported failures, traced the two numerical root causes, revised the implementation based on my own understanding, verified the reference values against mpmath, and ran the full test and formatting checks.
I reviewed the final code and take responsibility for it.

### DIFF
--- a/include/xsf/par_cyl.h
+++ b/include/xsf/par_cyl.h
@@ -1,7 +1,9 @@
 #pragma once
 
+#include "cephes/gamma.h"
 #include "error.h"
 #include "specfun/specfun.h"
+#include "trig.h"
 
 namespace xsf {
 namespace detail {
@@ -97,8 +99,8 @@ namespace detail {
         if (x < 0.0) {
             x1 = -x;
             vl = vvla(x1, va);
-            gl = specfun::gamma2(-va);
-            pd = pi * vl / gl + cos(pi * va) * pd;
+            gl = cephes::Gamma(-va);
+            pd = pi * vl / gl + cospi(va) * pd;
         }
         return pd;
     }
@@ -220,7 +222,7 @@ namespace detail {
         // ====================================================
 
         int ja, k, l, m, nk, nv, na;
-        T xa, vh, ep, f, f0, f1, v0, v1, v2, pd, pd0, pd1, s0;
+        T xa, vh, ep, f, f0, f1, v0, v1, v2, pd, pd0, pd1;
 
         xa = fabs(x);
         vh = v;
@@ -313,9 +315,8 @@ namespace detail {
                     f1 = f0;
                     f0 = f;
                 }
-                s0 = pd0 / f;
                 for (k = 0; k <= na; k++) {
-                    dv[k] *= s0;
+                    dv[k] = pd0 * (dv[k] / f);
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


verma-divyanshu-git:
@lucascolley, could you please take a look when you have time? this is the xsf fix for scipy/scipy#25814, and the scipy PR depends on it. thanks.

verma-divyanshu-git:
@lucascolley @steppi @j-bowhay, could one of you please review this when you have time? scipy/scipy#25966 is waiting on this change. the branch is mergeable, the full xsf suite passes (410 tests), and the scipy.special suite passes (8,769 tests). i'm happy to address any changes, if needed.

include/xsf/par_cyl.h
```diff
@@ -313,9 +315,8 @@ namespace detail {
                     f1 = f0;
                     f0 = f;
                 }
-                s0 = pd0 / f;
                 for (k = 0; k <= na; k++) {
-                    dv[k] *= s0;
+                    dv[k] = pd0 * (dv[k] / f);
```
dschmitz89, line 319: Was this a bug in the original specfun implementation?

include/xsf/par_cyl.h
```diff
@@ -97,8 +99,8 @@ namespace detail {
         if (x < 0.0) {
             x1 = -x;
             vl = vvla(x1, va);
-            gl = specfun::gamma2(-va);
-            pd = pi * vl / gl + cos(pi * va) * pd;
+            gl = cephes::Gamma(-va);
```
j-bowhay, line 102: Why is this change necessary?

j-bowhay:
I have some concerns about the use of AI here (AI fixing an AI issue :face_with_head_bandage: ) , a lot of these changes are quite subtle

include/xsf/par_cyl.h
verma-divyanshu-git: yes, this came from the old scipy specfun code before it was moved into xsf. for v=-1, x=50, pd0 / f is around 4.95e-418, so it becomes zero even though the final value still fits in a double. changing the order avoids that early underflow.

include/xsf/par_cyl.h
verma-divyanshu-git: my bad, this was left over from an earlier version of the fix and isn't needed anymore. i'll revert these lines and rerun the tests.

verma-divyanshu-git:
> I have some concerns about the use of AI here (AI fixing an AI issue 🤕 ) , a lot of these changes are quite subtle

i understand the concern. i did use AI assistance to map the pbdv call path, understand how the files fit together, and get an initial patch started.

but i reproduced the three failures, worked through the two separate causes, and revised the code based on that. i then checked the values and derivatives against 80-digit mpmath results and ran all 410 xsf tests plus 8,769 scipy.special tests. i can walk through every change.

PR #258, opened 2026-08-25
CI: try caching CuPy kernels

lucascolley:


lucascolley:
gonna merge to see if this works restoring from main, can revert if not

PR #260, opened 2026-08-26
CI: move GPU job caches to Cloudflare R2

aktech:
#### Reference issue
Follow up from gh-242 (https://github.com/scipy/xsf/pull/242#issuecomment-5411609987).

The on-prem GPU runner reaches the GitHub-hosted cache backend at ~14 MB/s, so restoring the 2.2 GB pixi environment took ~2.5 min of a ~4 min job. This moves both caches (pixi environment and CuPy kernels) to a Cloudflare R2 bucket via `cirunlabs/cache/s3`, and turns off setup-pixi's built-in cache so the environment is not fetched twice.

#### Additional information
Uses the `S3_ENDPOINT` / `S3_ACCESS_KEY_ID` / `S3_SECRET_ACCESS_KEY` repository secrets. Bucket: `scipy-cirun-github-actions-cache`. Docs: https://docs.cirun.io/caching/s3-compatible

#### AI Generation Disclosure
Claude Code (Opus) was used to draft the workflow changes. I take full responsibility of the changes.

### DIFF
--- a/.github/workflows/gpu_tests.yml
+++ b/.github/workflows/gpu_tests.yml
@@ -40,6 +40,7 @@ jobs:
           s3-access-key-id: ${{ secrets.S3_ACCESS_KEY_ID }}
           s3-secret-access-key: ${{ secrets.S3_SECRET_ACCESS_KEY }}
           s3-prefix: xsf
+          s3-ttl-days: 0  # expiry is a lifecycle rule on the bucket itself
 
       # Replaces setup-pixi's built-in cache (which is hardwired to the GitHub
       # cache backend). setup-pixi caches the `.pixi` directory next to
@@ -58,6 +59,7 @@ jobs:
           s3-access-key-id: ${{ secrets.S3_ACCESS_KEY_ID }}
           s3-secret-access-key: ${{ secrets.S3_SECRET_ACCESS_KEY }}
           s3-prefix: xsf
+          s3-ttl-days: 0  # expiry is a lifecycle rule on the bucket itself
 
       - name: run nvidia-smi
         run: nvidia-smi
--- a/.github/workflows/gpu_tests.yml
+++ b/.github/workflows/gpu_tests.yml
@@ -24,12 +24,40 @@ jobs:
           # We don't want to change the directory, and $HOME isn't available for `env:`
           echo "CUPY_CACHE_DIR=$HOME/.cupy/kernel_cache" >> $GITHUB_ENV
 
+      # The on-prem GPU runner has no fast path to the GitHub-hosted cache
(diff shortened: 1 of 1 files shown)


.github/workflows/gpu_tests.yml
```diff
@@ -13,6 +13,14 @@ jobs:
   cupy_tests:
     name: CuPy GPU
     runs-on: "cirun-scipy-onprem-gpu--${{ github.run_id }}"
+    env:
+      # The on-prem GPU runner has no fast path to the GitHub-hosted cache
+      # backend (restores ran at ~14 MB/s), so caches live in a Cloudflare R2
+      # bucket instead, via cirunlabs/cache/s3 (same interface as actions/cache,
+      # see https://docs.cirun.io/caching/s3-compatible). Repository secrets are
+      # not available to pull requests from forks, so those runs fall back to
+      # the GitHub cache below.
```
lucascolley, line 22: Hmmm, this seems like a very serious limitation. Probably we care about job runtime _more_ on PRs from forks

.github/workflows/gpu_tests.yml
```diff
@@ -24,13 +32,49 @@ jobs:
...
+      # cache backend). setup-pixi caches the `.pixi` directory next to
+      # pixi.lock; the key hashes the lockfile plus this workflow file, since
+      # the pixi version and environment selection live here and change what
+      # ends up in `.pixi`.
+      - name: Cache pixi environment (R2)
+        if: env.S3_CACHE == 'true'
+        uses: cirunlabs/cache/s3@0d5005be5be8cb3f20e624125e3fca0dc68eb384 # v5.0.5-cirun.1
+        with:
+          path: .pixi
+          key: gpu-pixi-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('pixi.lock', '.github/workflows/gpu_tests.yml') }}
```
pavelzw, line 68: technically it would be good to include the full path to `.pixi` in here since conda environments are not relocatable.
this would then also mirror what setup-pixi's cache is doing: https://github.com/prefix-dev/setup-pixi/blob/f00437f565399d418b0acc85936d12c1fb668347/src/cache.ts#L23-L44

.github/workflows/gpu_tests.yml
lucascolley: @aktech please could we rework this to have a publicly-readable bucket that PRs from forks can restore from without secrets, and only write to the cache on `main`? Is that something that cirun supports?

If so, we could get rid of the GitHub cache fallback, I think.

aktech: I'll explore that route.

PR #262, opened 2026-08-26
BUG(specfun): chguit: remove endpoint singularity

seanbreeden:
Addresses part of scipy/scipy#2287 (`hyperu` scipy vs. mpmath), open since 2012.

## Diagnosis

`chguit` evaluates DLMF 13.4.4 by composite Gauss-Legendre:

```
U(a,b,x) = 1/Γ(a) ∫₀^∞ e^{-xt} t^{a-1} (1+t)^{b-a-1} dt
```

For non-integer `a` the factor `t^{a-1}` is an algebraic endpoint singularity at
`t = 0`: continuous, but with unbounded derivatives. Gauss-Legendre converges
only algebraically against such a factor.

Three observations pin this down:

1. **The loop never converges.** Instrumenting the first panel loop shows it
   exhausting `for (m = 10; m <= 100; m += 5)` without its `1e-9` test ever
   firing, then returning the last iterate.
2. **The failure is silent.** `chguit` opens with `*id = 9;`, a constant rather
   than an estimate, so the caller is told 9 digits were achieved regardless.
   `chgu` only raises `isfer` when `id < 6`, so nothing flags.
3. **It is not a step-count problem.** Raising the cap from 100 to 2000 gains a
   factor of 7 and still leaves 1e-7 relative error.

The error also vanishes exactly at `a = 1`, where `t^0` has no singularity,
which is consistent with the diagnosis.

## Fix

Substituting `t = s^{1/a}` maps `t^{a-1} dt` to `(1/a) ds` exactly, leaving a
smooth integrand.

The substitution is not free: it places `s^{1/a}` inside the exponential, itself
singular at `s = 0` when `a > 1`. It therefore pays only while the induced
singularity is weaker than the original, `1/a > a - 1`, i.e. `a < φ ≈ 1.618`.
Applying it unconditionally regresses **2965 of 4608** grid points, one of them
from `2.5e-18` to `3.5e-02`. It is gated at `a < 1.4`, inside that bound with
margin.

## Validation

Gate chosen on a deterministic grid, then validated on **6000 randomly sampled
`(a, b, x)` with a seed and ranges independent of the tuning grid**, against
mpmath at 40 digits:

| | |
|---|---|
| improved | 61 |
| unchanged | 5939 |
| **regressed** | **0** |
| largest improvement | 63039x (1.16e-07 → 1.84e-12) |
| runtime, 40000 calls | 3062 ms vs 3077 ms |

Every assertion in the existing `test_hyperu.cpp` passes unchanged, including
the gh-15650 reference values at their `1e-13` tolerance.

The added test uses mpmath reference values at 50 digits and a `1e-8` tolerance,
chosen so it fails on 4 of its 8 points before the change and passes on all 8
after, rather than being satisfied by either.

## Scope

This addresses the small-`a` quadrature path only. It does **not** fix the
separate defect in the 2019 comment on gh-2287, `hyperu(-90, 1, 10.0)` returning
`3.8e+143` against a true `-1.9e+139`; that takes a different branch and is
untouched here.

It also does not reach full double precision. Worst remaining error across the
new test points is `5.98e-10`, down from `3.94e-07`. The residual is likely the
`*id = 9` constant and the `1e-9` convergence target, both of which could be
tightened separately.

### DIFF
--- a/include/xsf/specfun/specfun.h
+++ b/include/xsf/specfun/specfun.h
@@ -1197,23 +1197,47 @@ namespace specfun {
         b1 = b - a - 1.0;
         c = 12.0 / x;
         hu0 = 0.0;
+        // The integrand carries t^(a-1), an algebraic endpoint singularity at
+        // t = 0 for non-integer a: continuous, unbounded derivatives, and
+        // Gauss-Legendre converges only algebraically against it.
+        //
+        // The substitution t = s^(1/a) maps t^(a-1) dt to (1/a) ds exactly, but
+        // it is not free: it puts s^(1/a) inside the exponential, which is
+        // itself singular at s = 0 when a > 1. It therefore pays only when the
+        // induced singularity is weaker than the original, i.e. 1/a > a - 1,
+        // which is a < golden ratio; 1.4 is used, inside that with margin.
+        // Outside it the substitution makes matters sharply
+        // worse (measured: applying it unconditionally regressed 2965 of 4608
+        // grid points, one of them from 2.5e-18 to 3.5e-02).
+        const bool use_sub = (a1 != floor(a1)) && (a < 1.4) && (a > 0.0);
+        const double hi = use_sub ? pow(c, a) : c;
         for (m = 10; m <= 100; m += 5) {
             hu1 = 0.0;
-            g = 0.5 * c / m;
+            g = 0.5 * hi / m;
             d = g;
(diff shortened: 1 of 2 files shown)


fbourgey:
@seanbreeden thanks. Which AI tools did you use for this (if any)? Pretty sure you did as the description looks like it was AI-generated. Why would you remove the AI disclosure?

PR #264, opened 2026-09-02
Let poisson_binom distribution function kernels take generic integers

steppi:

In preparation for making `poisson_binom_cdf` public in SciPy as I proposed here, https://github.com/scipy/scipy/issues/25921#issuecomment-5287134536, this updates the `take_from_discrete_cdf` and `take_from_pmf` functions to take template args for the integer type so that they will work for generic integers. To do this in a streamlined way I added in implementation of the [`std::cmp_less`](https://en.cppreference.com/cpp/utility/intcmp) template function for comparing integers of arbitrary types. Although this cannot be used in CuPy until github.com/cupy/cupy/pull/9959 lands, I've gone ahead and added the relevant helper templates to `config.h`, along with some related helpers which aren't needed here but seem like they could be desirable in the future when we move to supporting generic integer type across more functions.

#### AI Generation Disclosure

I've used GPT 5.6 Sol to help review my diffs locally

### DIFF
--- a/include/xsf/config.h
+++ b/include/xsf/config.h
@@ -222,6 +222,24 @@ XSF_HOST_DEVICE complex<T> pow(const complex<T> &x, const T &y) {
 template <typename T>
 using is_floating_point = cuda::std::is_floating_point<T>;
 
+template <typename T>
+using is_integral = cuda::std::is_integral<T>;
+
+template <typename T>
+inline constexpr bool is_integral_v = cuda::std::is_integral_v<T>;
+
+template <typename T>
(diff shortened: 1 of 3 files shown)


steppi:
I can't run GPU tests  locally now because my power is out and my GPU workstation doesn't have a battery backup. Hopefully the CI job passes.

steppi:
The GPU CI failure is unrelated. The `xsref` repo couldn't be accessed for some reason. I see transient GitHub failures like this pretty often now, so I'm not sure if there is anything wrong on our end. I have power again and CuPy tests pass locally for me, so I think this should be fine to merge.

PR #266, opened 2026-09-04
TST: run CuPy tests in parallel

fbourgey:
Let us see if CI allows us to run CuPy tests in parallel. The tests were much faster locally.

#### AI Generation Disclosure

Asked Codex for feedback.


lucascolley (approved):
Oh yeah I should have caught this already, thanks Florian!

steppi:
I'm getting deprecation warnings when running this locally

```
python_tests/test_cupy.py::TestCuPy::test_erfc[tables_paths2]
  /home/birbir/xsf/.pixi/envs/cupy-tests/lib/python3.14/multiprocessing/popen_fork.py:76: DeprecationWarning: This process (pid=150856) is multi-threaded, use of fork() may lead to deadlocks in the child.
    self.pid = os.fork()
```

also, since the only step worth giving CPU parallelization to is kernel compilation, perhaps it makes more sense to have a parallelized function to build/rebuild all of the kernels (that need rebuilding) and warm the cache before running the tests

PR #267, opened 2026-09-04
TST: clean up CuPy test definitions

fbourgey:
- Fix the mismatch for `expn` https://github.com/scipy/xsf/blob/main/include/xsf/cephes/expn.h#L147 for CuPy tests (changing to `n,x`)

- `wofz` only accepts complex arguments https://github.com/scipy/xsf/blob/main/include/xsf/erf.h#L78 so this PR removes the signature `'f->f', 'd->d'`

- Some cleaning (unused imports and whitepace)

#### AI Generation Disclosure

Worked with Codex on this.

### DIFF
--- a/python_tests/test_cupy.py
+++ b/python_tests/test_cupy.py
@@ -10,8 +10,6 @@ import os
 import numpy as np
 import polars as pl
 import pytest
-import shutil
-import tempfile
 
 from packaging.version import Version
 from pathlib import Path
@@ -228,13 +226,13 @@ class TestCuPy:
         input_path, output_path, tol_path = tables_paths
         expn = cupy._core.create_ufunc(
             'cupyx_scipy_special_expn',
-            ('ff->f', 'dd->d'),
+            ('ld->d', 'ff->f', 'dd->d'),
             'out0 = xsf::cephes::expn(in0, in1)',
             preamble=get_preamble("xsf/cephes/expn.h"),
         )
 
-        x, n = get_cols_as_cupy(input_path)
-        out = cupy.asnumpy(expn(x, n))
+        n, x = get_cols_as_cupy(input_path)
+        out = cupy.asnumpy(expn(n, x))
 
         desired = get_cols_as_numpy(output_path)
         rtol = get_cols_as_numpy(tol_path)
@@ -297,7 +295,7 @@ class TestCuPy:
         input_path, output_path, tol_path = tables_paths
         _lambertw_scalar = cupy._core.create_ufunc(
             "cupyx_scipy_lambertw_scalar",
-            ("Dld->D", "Fif->f"),
+            ("Dld->D", "Flf->F"),
             "out0 = xsf::lambertw(in0, in1, in2)",
             preamble=get_preamble("xsf/lambertw.h"),
         )
@@ -356,7 +354,7 @@ class TestCuPy:
             ),
             preamble=get_preamble("xsf/sici.h"),
         )
-            
+
         x = get_cols_as_cupy(input_path)
         if cupy.iscomplexobj(x):
             pytest.xfail("Known bug, returning nan instead of a complex infinity.")
@@ -364,8 +362,6 @@ class TestCuPy:
 
         desired0, desired1 = get_cols_as_numpy(output_path)
         rtol0, rtol1 = get_cols_as_numpy(tol_path)
-        error = extended_relative_error(out0, desired0)
-        tol = self._adjust_tol(rtol0)
         assert np.all(
             extended_relative_error(out0, desired0) <= self._adjust_tol(rtol0)
         )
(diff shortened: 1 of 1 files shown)


python_tests/test_cupy.py
```diff
@@ -228,13 +226,13 @@ def test_expn(self, tables_paths):
         input_path, output_path, tol_path = tables_paths
         expn = cupy._core.create_ufunc(
             'cupyx_scipy_special_expn',
-            ('ff->f', 'dd->d'),
+            ('ld->d', 'ff->f', 'dd->d'),
```
steppi: The previous signature matches what CuPy has had prior to `xsf`

https://github.com/cupy/cupy/blob/d7c8f0cdcda22a850aa9ae75453e45acbadef58e/cupyx/scipy/special/_expn.py#L337-L361

Their kernel has the same types as ours too

https://github.com/cupy/cupy/blob/d7c8f0cdcda22a850aa9ae75453e45acbadef58e/cupyx/scipy/special/_expn.py#L220

I think most of these curious signatures you found were copied directly from CuPy. They don't have the same concerns regarding portability of integer types because supported GPUs give a more uniform set of platforms.

Some of these things are genuinely odd, but I'd suggest taking this upstream to CuPy rather than changing our tests to drift from what is in CuPy. It seems worth investigating why CuPy has things the way they do first though.

fbourgey: Thanks, reverted. I'll try to investigate. 

What do you think of the other changes in this PR?

steppi: The `wofz` change seems fine. That's not even in CuPy yet, so I think I just copied that signature from another test. It seems there was no follow up on https://github.com/cupy/cupy/issues/9907 yet, so maybe one of us should go ahead and add those functions to CuPy.

The cosmetic fixups look good too. I'd be happy to merge this after the PR title is updated to reflect the content that's here now.

fbourgey: I updated the PR title. I can try to open a PR adding `wofz` to CuPy.

steppi: Thanks @fbourgey!

steppi (approved):
Looks good to me

fbourgey:
@steppi, I see that you added `lambertw` to CuPy in https://github.com/cupy/cupy/pull/8140 with a ufunc signature `("Dld->D", "Fif->f")`. Can you explain why this is not `("Dld->D", "Flf->F")`?

steppi:
> @steppi, I see that you added `lambertw` to CuPy in [cupy/cupy#8140](https://github.com/cupy/cupy/pull/8140) with a ufunc signature `("Dld->D", "Fif->f")`. Can you explain why this is not `("Dld->D", "Flf->F")`?

Many GPUs don't have good support for all 64 bit types, so an all 32 bit signature is desirable. We should offer all three through.

fbourgey:
> > @steppi, I see that you added `lambertw` to CuPy in [cupy/cupy#8140](https://github.com/cupy/cupy/pull/8140) with a ufunc signature `("Dld->D", "Fif->f")`. Can you explain why this is not `("Dld->D", "Flf->F")`?
> 
> Many GPUs don't have good support for all 64 bit types, so an all 32 bit signature is desirable. We should offer all three through.

Right but the output is always complex, no? So something like `("Dld->D", "Flf->F", "Fif->F")`?

https://github.com/scipy/xsf/blob/main/include/xsf/lambertw.h#L533

steppi:
> > > @steppi, I see that you added `lambertw` to CuPy in [cupy/cupy#8140](https://github.com/cupy/cupy/pull/8140) with a ufunc signature `("Dld->D", "Fif->f")`. Can you explain why this is not `("Dld->D", "Flf->F")`?
> > 
> > 
> > Many GPUs don't have good support for all 64 bit types, so an all 32 bit signature is desirable. We should offer all three through.
> 
> Right but the output is always complex, no? So something like `("Dld->D", "Flf->F", "Fif->F")`?
> 
> https://github.com/scipy/xsf/blob/main/include/xsf/lambertw.h#L533

wait, is it really lower case f? that's just a typo. i'm surprise that made it in then

fbourgey:
I think so https://github.com/cupy/cupy/blob/main/cupyx/scipy/special/_lambertw.py#L12

steppi:
> I think so https://github.com/cupy/cupy/blob/main/cupyx/scipy/special/_lambertw.py#L12

Wow, that's crazy. That should be fixed

fbourgey:
I'll open a PR over at CuPy and suggest the three `("Dld->D", "Flf->F", "Fif->F")`

steppi:
> I'll open a PR over at CuPy and suggest the three `("Dld->D", "Flf->F", "Fif->F")`

We actually don't even have a kernel capable of providing the `FiF->F` loop though, so can leave it out for now. Nor do we have actual complex64 support.  I'd like to add more single precision support everywhere in `xsf`

fbourgey:
are you suggesting we do nothing at CuPy now?

steppi:
Taking a look at what we actually offer, I would change it to `Flf->F` and `Dld->D` in CuPy like you suggested originally

python_tests/test_cupy.py
fbourgey: I opened a PR for wofz https://github.com/cupy/cupy/pull/10276

PR #268, opened 2026-09-04
ENH: add geninvgauss and genhyperbolic kernels

fbourgey:
Add logpdf and pdf kernels for the generalized inverse Gaussian and generalized hyperbolic. They look like natural candidates for xsf.

Those are direct C++ translations of the Cython code

https://github.com/scipy/scipy/blob/main/scipy/stats/_stats.pyx#L602
https://github.com/scipy/scipy/blob/main/scipy/stats/_stats.pyx#L434

I added permalinks to SciPy tests.

#### AI Generation Disclosure

Worked alongside Codex for this, notably for porting the SciPy tests to XSF.

### DIFF
--- a/include/xsf/cpu/stats.h
+++ b/include/xsf/cpu/stats.h
@@ -210,7 +210,7 @@ namespace cpu {
         double t1, t2, t3, t4, t5;
 
         t1 = detail::genhyperbolic_log_norming_constant(p, a, b);
-        t2 = std::sqrt(1.0 + x * x);
+        t2 = std::hypot(1, x);
         t3 = (p - 0.5) * std::log(t2);
         t4 = std::log(cyl_bessel_ke(p - 0.5, a * t2)) - a * t2;
         t5 = b * x;
--- a/include/xsf/cpu/stats.h
+++ b/include/xsf/cpu/stats.h
@@ -159,5 +159,81 @@ namespace cpu {
         return std::min(std::max(cdf, 0.0), 1.0);
     }
 
+    // Logarithm of the generalized inverse Gaussian probability density function.
+    inline double geninvgauss_logpdf(double x, double p, double b) {
+        if (x <= 0) {
+            return -std::numeric_limits<double>::infinity();
+        }
+
+        double z = cyl_bessel_ke(p, b);
+        if (std::isinf(z)) {
+            return std::numeric_limits<double>::quiet_NaN();
+        }
+
+        return -std::log(2.0) - std::log(z) + b + (p - 1.0) * std::log(x) - b * (x + 1.0 / x) / 2.0;
+    }
+
+    inline float geninvgauss_logpdf(float x, float p, float b) {
+        return static_cast<float>(
+            geninvgauss_logpdf(static_cast<double>(x), static_cast<double>(p), static_cast<double>(b))
+        );
+    }
+
+    // Generalized inverse Gaussian probability density function.
+    inline double geninvgauss_pdf(double x, double p, double b) { return std::exp(geninvgauss_logpdf(x, p, b)); }
+
+    inline float geninvgauss_pdf(float x, float p, float b) {
+        return static_cast<float>(
+            geninvgauss_pdf(static_cast<double>(x), static_cast<double>(p), static_cast<double>(b))
+        );
+    }
+
+    namespace detail {
+
+        inline double genhyperbolic_log_norming_constant(double p, double a, double b) {
+            double t1, t2, t3, t4, t5, t6;
+
+            t1 = (a + b) * (a - b);
+            t2 = p * 0.5 * std::log(t1);
+            t3 = 0.5 * std::log(2 * M_PI);
+            t4 = (p - 0.5) * std::log(a);
+            t5 = std::sqrt(t1);
+            t6 = std::log(cyl_bessel_ke(p, t5)) - t5;
+
+            return t2 - t3 - t4 - t6;
+        }
+
+    } // namespace detail
+
+    // Logarithm of the generalized hyperbolic probability density function.
+    inline double genhyperbolic_logpdf(double x, double p, double a, double b) {
+        double t1, t2, t3, t4, t5;
+
+        t1 = detail::genhyperbolic_log_norming_constant(p, a, b);
+        t2 = std::sqrt(1.0 + x * x);
+        t3 = (p - 0.5) * std::log(t2);
+        t4 = std::log(cyl_bessel_ke(p - 0.5, a * t2)) - a * t2;
+        t5 = b * x;
+
+        return t1 + t3 + t4 + t5;
+    }
+
+    inline float genhyperbolic_logpdf(float x, float p, float a, float b) {
+        return static_cast<float>(genhyperbolic_logpdf(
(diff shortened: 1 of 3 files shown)


include/xsf/cpu/stats.h
```diff
@@ -159,5 +159,81 @@ namespace cpu {
...
+        }
+
+    } // namespace detail
+
+    // Logarithm of the generalized hyperbolic probability density function.
+    inline double genhyperbolic_logpdf(double x, double p, double a, double b) {
+        double t1, t2, t3, t4, t5;
+
+        t1 = detail::genhyperbolic_log_norming_constant(p, a, b);
+        t2 = std::sqrt(1.0 + x * x);
```
dschmitz89:
```suggestion
        t2 = std::hypot(1, x);
```

dschmitz89 (approved):

dschmitz89:
I think those functions are called as a `LowLevelCallable` for the CDFs. Does our infra enable that for functions exposed in `special` as well?

fbourgey:
> I think those functions are called as a `LowLevelCallable` for the CDFs. Does our infra enable that for functions exposed in `special` as well?

Hmm good catch. I do not think our current infra allows it. cc @steppi @mdhaber to confirm

dschmitz89:
> > I think those functions are called as a `LowLevelCallable` for the CDFs. Does our infra enable that for functions exposed in `special` as well?
> 
> Hmm good catch. I do not think our current infra allows it. cc @steppi @mdhaber to confirm

[This](https://github.com/scipy/scipy/blob/b272202d298cf6634a404c3930669e61501aa1c0/scipy/stats/_continuous_distns.py#L4146) the relevant part for `genhyperbolic` in SciPy. We basically need to expose the function as straight forward C functions and refer them correctly from inside python from what I understand. I am sure that it is possible but this stuff is hairy.

@steppi @WarrenWeckesser Could you weigh in?

I propose to still merge this PR in the meanwhile. Finding the correct way to plumb these functions into SciPy infra is decooupled from an initial correct implementation of the functions themselves. A different interface can be added later on. A PR to SciPy will reveal what actually needs to be done. WDYT @fbourgey ?

fbourgey:
> > > I think those functions are called as a `LowLevelCallable` for the CDFs. Does our infra enable that for functions exposed in `special` as well?
> > 
> > 
> > Hmm good catch. I do not think our current infra allows it. cc @steppi @mdhaber to confirm
> 
> [This](https://github.com/scipy/scipy/blob/b272202d298cf6634a404c3930669e61501aa1c0/scipy/stats/_continuous_distns.py#L4146) the relevant part for `genhyperbolic` in SciPy. We basically need to expose the function as straight forward C functions and refer them correctly from inside python from what I understand. I am sure that it is possible but this stuff is hairy.
> 
> @steppi @WarrenWeckesser Could you weigh in?
> 
> I propose to still merge this PR in the meanwhile. Finding the correct way to plumb these functions into SciPy infra is decooupled from an initial correct implementation of the functions themselves. A different interface can be added later on. A PR to SciPy will reveal what actually needs to be done. WDYT @fbourgey ?

I agree with you.

steppi:
If these aren't going to be used as ufunc kernels, but only in SciPy for low level callables that are passed into `integrate.quad`, then I'm not sure they are a good fit for `xsf`. I envision `xsf` as a library for portable ufunc and gufunc kernels. A ufunc that performs the numerical integration to compute `_integrate_pdf` could be a good fit for xsf, Perhaps the ufuncs for the kernels given here could be used with `integrate.tanhsinh` or something instead of `integrate.quad` or something to give an array API implementation of `_integrate_pdf` here. What's the plan for even exposing alternative backend support for that though?

mdhaber:
I don't think they would be used for `LowLevelCallable`s to pass into `quad`. There's less reason to these days, e.g.

<details>

```python3
import numpy as np
from scipy import stats, integrate, special

def f(x, p, b):
    return x**(p-1) * np.exp(-b * (x+1/x)/2) / (2 * special.kv(p, b))

p, b = 2.3, 1.5
x = np.linspace(0, 10, 200)

ref = stats.geninvgauss.cdf(x, p, b)
res = integrate.tanhsinh(lambda x: f(x, p, b), 0, x)
stats.gmean(np.abs(res.integral - ref)[1:])
# np.float64(2.6682833680564e-13)
```

</details>

```
%timeit stats.geninvgauss.cdf(x, p, b)
# 18.6 ms ± 2.48 ms per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit integrate.tanhsinh(lambda x: f(x, p, b), 0, x)
# 3.77 ms ± 92.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

But it might be nice to have these available as ufuncs for evaluating the PDF and as the integrand for calculating the CDF. Even better if `xsf` could take care of the integration.

steppi:
Those are some very nice results from `tanhsinh` @mdhaber, assuming the accuracy is comparable. I see no reason to block this PR merging this PR at least.

mdhaber:
Yeah, the `gmean` of the absolute error is in the details (`2.6682833680564e-13`). I haven't tested against arbitary precision to see which is more accurate, but I think we've tested `tanhsinh` quite a bit and it's typically comparable or better as long as there is not a singularity on the right.

dschmitz89:
So we could merge this PR and replace the low level callable/quad integration with tanhsinh of the ufunc over at SciPy to avoid performance regressions.

mdhaber:
I'd prefer that the old distributions not use `tanhsinh`. Is suggest that the old can remain old, and those who want benefits of the new can use the new. (I won't raise a fuss if you want to use `cubature`, though.)

One might say that we don't have a new replacement for `geninvgauss` and `genhyperbolic` (except via `make_distribution`) and want to create them. I wish that sort of effort would be directed towards https://github.com/statistical-python/scikit-stats, where (ideally) distributions would be created and maintained by the folks who care about them. `scipy.stats` has enough going on besides maintaining 100+ distributions, and there needs to be a place for distributions that aren't accepted into SciPy for whatever reason. (I could really use some help setting up the infrastructure there, BTW. I suppose I could enlist AI these days but I haven't gotten around to it.)

Of course, that's just my opinion.

dschmitz89:
> I'd prefer that the old distributions not use `tanhsinh`. Is suggest that the old can remain old, and those who want benefits of the new can use the new. (I won't raise a fuss if you want to use `cubature`, though.)
> 
> One might say that we don't have a new replacement for `geninvgauss` and `genhyperbolic` (except via `make_distribution`) and want to create them. I wish that sort of effort would be directed towards https://github.com/statistical-python/scikit-stats, where (ideally) distributions would be created and maintained by the folks who care about them. `scipy.stats` has enough going on besides maintaining 100+ distributions, and there needs to be a place for distributions that aren't accepted into SciPy for whatever reason. (I could really use some help setting up the infrastructure there, BTW. I suppose I could enlist AI these days but I haven't gotten around to it.)
> 
> Of course, that's just my opinion.

Bringing `scikit-stats` up to speed sounds like a good long term plan. How can we move forward with getting rid of the pdfs implemented in cython for SciPy though? Just leave them be for the time being?

mdhaber:
Oh when you put it that way I suppose I'm on board. I would like to remove the Cython files. How about `cubature` for the integration : )  that may be faster even for scalars.

PR #270, opened 2026-09-10
ENH: add trivariate normal cdf

fbourgey:
#### Reference issue
https://github.com/scipy/xsf/issues/98

This took me some time but I think this is ready for review. This PR adds the trivariate normal CDF. This is a direct translation of Alan Genz's MATLAB code https://web.archive.org/web/20200205123040/http://www.math.wsu.edu/faculty/genz/software/matlab/tvn.m

- The license is now in `LICENSES_bundle.txt` instead of referring to the link
- Allows tiny negative input determinants (8 x machine precision) but this is up to discussion.
- It returns exactly zero when r23==-1 and h2<=-h3 while (I think) MATLAB does not.

#### Additional information
while working on this, I think it would be best to only implement CDFs. Looking at the linked issue, more are to come.. As of now, we have the SF for the bivariate normal. I could do the change now or later.

#### AI Generation Disclosure

Worked in pair with Codex on this, notably on the tests and the (small) differences with the original MATLAB code.


fbourgey:
Moving this to draft now. It is best to wait for https://github.com/scipy/xsf/pull/273 to get merged.

include/xsf/multivariate_normal.h
```diff
@@ -189,4 +189,243 @@ XSF_HOST_DEVICE inline float bivariate_normal_cdf(float dh, float dk, float r) {
     return bivariate_normal_cdf(static_cast<double>(dh), static_cast<double>(dk), static_cast<double>(r));
 }
 
+namespace detail {
+
+    XSF_HOST_DEVICE inline void sincs(double x, double &sx, double &cs) {
```
dschmitz89: Do we loose accuracy if we replace this at all call sites with plain `std::sin(x)` and `std::pow(std::cos(x), 2)`? Can also be done in a follow up though.

include/xsf/multivariate_normal.h
```diff
@@ -189,4 +189,243 @@ XSF_HOST_DEVICE inline float bivariate_normal_cdf(float dh, float dk, float r) {
...
+        }
+
+        double t = wid * xgk[10];
+        fc = tvnf(cen - t, h1, h2, h3, r23, a12, a13) + tvnf(cen + t, h1, h2, h3, r23, a12, a13);
+        resk = wid * (resk + wgk[10] * fc);
+        err = std::abs(resk - wid * resg);
+    }
+
+    XSF_HOST_DEVICE inline double
+    adonet(double h1, double h2, double h3, double r23, double a12, double a13, double tol) {
```
dschmitz89: More descriptive name would be?

dschmitz89 (approved):
I believe that this is a faithful translation of the old code. Review comments are as always with such codes about future maintainability: is the helper function for sin and cos^2 actually needed (seems like over optimization to me) and can the functions have more descriptive names?

include/xsf/multivariate_normal.h
fbourgey: Thanks. I'll investigate and raise an issue for a follow-up PR.

fbourgey:
@dschmitz89 thanks for the review. I am now using more explicit function names.

dschmitz89 (approved):

PR #273, opened 2026-09-11
MAINT: use CDF instead of SF for bivariate normal

fbourgey:
As I worked on https://github.com/scipy/xsf/pull/270, it appeared clear (at least to me) that we should only have CDFs. This PR replaces the existing the SF implementation with CDF for the bivariate normal distribution.

#### AI Generation Disclosure

Worked with Codex.

### DIFF
--- a/include/xsf/multivariate_normal.h
+++ b/include/xsf/multivariate_normal.h
@@ -53,19 +53,22 @@ namespace detail {
 
 } // namespace detail
 
-XSF_HOST_DEVICE inline double bivariate_normal_sf(double dh, double dk, double r) {
-    // Survival function of the standard bivariate normal distribution.
+XSF_HOST_DEVICE inline double bivariate_normal_cdf(double dh, double dk, double r) {
+    // Cumulative distribution function of the standard bivariate normal distribution.
     //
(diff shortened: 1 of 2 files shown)


include/xsf/multivariate_normal.h
```diff
@@ -53,19 +53,22 @@ namespace detail {
...
     //
-    // Return P(X > dh, Y > dk) for a standard bivariate normal vector
+    // Return P(X <= dh, Y <= dk) for a standard bivariate normal vector
     // (X, Y) with correlation r.
     //
-    // dh, dk are the lower limits of the upper tail, and r must satisfy
+    // dh, dk are the upper limits of the lower tail, and r must satisfy
     // -1 <= r <= 1.
     //
     // Adapted from the MATLAB original implementation by Dr. Alan Genz;
```
dschmitz89: Could you add a comment that Genz had implemented the survival function? That would make it easier to understand why we for example have `detail::bivariate_normal_sf_boundary` later on.

dschmitz89 (approved):
LGTM only one comment about adding additional comments (pun intended).

fbourgey:
> LGTM only one comment about adding additional comments (pun intended).

Done :) I've also added a permalink.

PR #276, opened 2026-09-14
MAINT(airy): airyb: convert to pass-by-reference signature

### LINKED ISSUE #61: MAINT(airy): airyb: inconsistent signature
All functions in the `xsf` namespace that produce multiple scalar outputs are "returned" using the fancy C++ [pass-by-reference](https://www.ibm.com/docs/en/zos/2.4.0?topic=calls-pass-by-reference-c-only) style, except for `airyb`. For [xsf-rust](https://github.com/jorenham/xsf-rust) that's mildy annoying, because it requires special casing. But I can imagine that I'm not the only one that got confused when seeing this.

---

https://github.com/scipy/xsf/blob/9659b92ad3dcbc859f9d96d0cded2042aa4bedb0/include/xsf/airy.h#L135

I would expect it to look more like its `airy`-senpai (or any other multi-output scalar function in the main `xsf` namespace for that matter):

https://github.com/scipy/xsf/blob/9659b92ad3dcbc859f9d96d0cded2042aa4bedb0/include/xsf/airy.h#L442-L443

---

Changing this would obviously be backwards incompatible. So the question is; does that matter? For `xsf-rust` and `scipy`, `xsf` is a git submodule, so they'll probably survive. But I'm not sure about the other ways in which `xsf` might be used.

lucascolley:
closes gh-61 @jorenham 

I checked that neither SciPy nor CuPy nor xsf-rust calls this function, so this should be good to go.


jorenham:
Thanks for considering xsf-rust. But there's no need to worry (too much) about breaking xsf-rust, because it bundles a specific `xsf` version (xsf is a git submodule) :).

steppi (approved):
Thanks @lucascolley! I guess I missed this one when converting pointers to references for out args

Issue #278, opened 2026-09-18
BLD/TST: CuPy tests can get stuck on a stale version of xsf headers due to current pixi set up


steppi:
I noticed this while trying to figure out why some tests were passing the `xsf`` CuPy suite for me locally on a particular development branch but all CuPy tests for functions with xsf-backed kernels were failing when I used that feature branch in CuPy's xsf submodule.

This line here

https://github.com/scipy/xsf/blob/d06a5b0b9f2f9b2d0a4dfcfba3a17363200a66de/pixi.toml#L78

treats `xsf` as a source dependency. This causes `pixi` to build it separately and install as a CONDA package.

we see here how the tests were updated to pull the headers from this CONDA package

https://github.com/scipy/xsf/blob/d06a5b0b9f2f9b2d0a4dfcfba3a17363200a66de/python_tests/test_cupy.py#L84

but `xsf` is a header only library and CuPy JIT compiles directly out of the headers, so there are no build artifacts that can be used to detect if a change has been made.

@lucascolley, we could go back to something like what I was doing here

https://github.com/scipy/xsf/commit/0169ac50cda3188c98889bf8883b49270e73182d

but I figured you might know a better way to handle this.

lucascolley:
could you say more about the situation in which it ends up testing a stale version?

lucascolley:
any change to the xsf headers should cause a rebuild of xsf inside the `cupy-tests` environment. Docs at https://pixi.prefix.dev/latest/build/backends/pixi-build-cmake/#input-tracking

steppi:
> could you say more about the situation in which it ends up testing a stale version?

I cloned xsf on my Linux workstation, ran `pixi run test-cupy`. pixi creates the environment and runs the tests. If I then edit the headers in my working tree and even commit my work, so long as I reuse the same environment, the original headers will still get used when I run `pixi run test-cupy` again. If I run `git clean -fxd`, running `pixi run test-cupy` will cause the environment to be rebuilt (since it will be removed), and then the updated header will be used.

steppi:
> any change to the xsf headers should cause a rebuild of xsf inside the `cupy-tests` environment. Docs at https://pixi.prefix.dev/latest/build/backends/pixi-build-cmake/#input-tracking

[...]

PR #279, opened 2026-09-18
BUG/TST: Use xsf path directly in working tree for CuPy tests

steppi:
#### Reference issue

Refers to #278, but leaving that one open until a principled fix is found

This goes back to the old way of finding the xsf include path for the CuPy tests. It just grabs them directly from
`xsf/include` in the working tree. #202 had changed this to using an isolated install of the `xsf` Conda package, but it seems `pixi-build-cmake` is not successfully tracking changes in the headers in at least some cases, so they can become stale.

I've added a TODO comment that makes it clear this is only a temporary fix.
#### Additional information

Do you think it's fine to do this as a temporary fix @lucascolley?

#### AI Generation Disclosure

no AI


lucascolley (approved):

steppi:
Thanks!

PR #280, opened 2026-09-20
BLD: fix input tracking for header-only package

### LINKED ISSUE #278: BLD/TST: CuPy tests can get stuck on a stale version of xsf headers due to current pixi set up

lucascolley:
closes gh-278


steppi:
Thanks Lucas. I've confirmed this works. 

FYI, there's another issue that can cause stale kernels to get tested. I noticed in #258, that you removed the section I had added to avoid caching the kernels. CuPy doesn't check for updates in the headers though, so it will happily reuse the old kernels after the headers are changed. Lately I've found myself having to manually delete the `kernel_cache` when testing locally. That's a separate thing though, so I'm happy to merge this PR now.

steppi (approved):
Looks good to me

lucascolley:
> FYI, there's another issue that can cause stale kernels to get tested. I noticed in #258, that you removed the section I had added to avoid caching the kernels. CuPy doesn't check for updates in the headers though, so it will happily reuse the old kernels after the headers are changed.

ah, sorry, I assumed the problem was that it was causing crashes rather than silent caching issues.

> Lately I've found myself having to manually delete the `kernel_cache` when testing locally. 

How about incorporating an `rm` call into the Pixi task? Should be pretty simple.

steppi:
> > FYI, there's another issue that can cause stale kernels to get tested. I noticed in #258, that you removed the section I had added to avoid caching the kernels. CuPy doesn't check for updates in the headers though, so it will happily reuse the old kernels after the headers are changed.
> 
> ah, sorry, I assumed the problem was that it was causing crashes rather than silent caching issues.
> 
> > Lately I've found myself having to manually delete the `kernel_cache` when testing locally.
> 
> How about incorporating an `rm` call into the Pixi task? Should be pretty simple.

Yeah, that seems reasonable.

lucascolley:
Would this affect CI post cache-restore as well? If so, maybe we could investigate an alternative fix that has proper cache invalidation?

steppi:
> Would this affect CI post cache-restore as well? If so, maybe we could investigate an alternative fix that has proper cache invalidation?

Yes, I think this may actually affect CI. It seems like we need to track which headers have changed, including transitively through one of their includes

PR #281, opened 2026-09-20
DEV: enforce `__cuda` compatibility at the Pixi level

lucascolley:
hey @jepler, please could you try this branch?

The rough idea is that the `cupy-tests` environment is now restricted to machines which Pixi detects as supporting CUDA>=13 or CUDA>=12. If your machine has CUDA >=13 drivers, Pixi should choose the CUDA 13 environment, for which we include `cuda-version==13.0` as a dependency so that all packages have metadata compatible with all CUDA >=13 versions. Likewise, if you have CUDA >=12,<13 drivers, Pixi should choose the CUDA 12 environment, for which we include `cuda-version==12.0`.

If Pixi doesn't detect either of these, it should fail with a clear error message. It should refuse to run the cupy tests outright on a CPU-only `linux-64` machine.

If we ever require testing against specific minor versions of CUDA, we can introduce more platforms with higher minor cuda versions.


jepler:
```
HEAD is now at 3b11d2a DEV: enforce `__cuda` compatibility at the Pixi level
jepler@eric:~/src/xsf$ pixi run test-cupy
✨ Pixi task (test-cupy in cupy-tests): pytest -n auto --forked python_tests/test_cupy.py: (Run CuPy integration tests.)
Error:   × Cannot install environment 'cupy-tests': no platform supported by it matches the current system
```

You mentioned 12.x and 13.x, but my Debian system has nvidia-cuda-dev `11.8.89~11.8.0-5~deb12u1` as the installable version. So, the part where you reject old versions promptly looks correct.

It looks like I made a mistake when manually upgrading this system to trixie, with the "non-free" packages still coming from bookworm (oldstable). I'll fix this and try again once I have nvidia-cuda-dev 12.4.127 and other matching packages.

jepler:
With my local environment upgraded to cuda 12, the tests run to completion:
```
$ pixi run test-cupy
[snipped]
26 passed, 3 xfailed, 1 xpassed, 30 warnings in 109.16s (0:01:49)
```

lucascolley:
great, I'll merge this now to unblock you then, pretty sure it is correct

Issue #282, opened 2026-09-21
MAINT: use a portability namespace rather than injecting into `std`


steppi:
Early on in `xsf` development, when this codebase was still part of SciPy, @izaid proposed injecting functionality from `cuda::std` and `thrust` into the `std` namespace as an expedient to make development easier (see https://github.com/scipy/scipy/pull/19601/). We were warned multiple times not to do this, including here, https://github.com/scipy/scipy/pull/19601/#issuecomment-1834596405, by @jrhemstad the lead developer of CCCL,. He was of course right that this can lead to nasty problems, but so long as `xsf` was only used internally in SciPy and CuPy, and not intended for public consumption, injecting into `std` was workable.

We'd never intended to keep the current pattern around forever though, and now seems like a good time to move to a more principled solution, since `xsf` has a decent contributor-base now and we all hope to turn it into a useful library that is more widely used beyond its place as a dependency of SciPy and CuPy. 

Rather than injecting into `std`, we should instead use a portability namespace like `xsf::cxx`. This will be straightforward to accomplish, but will touch the entire codebase, and will change how all contributors must write their code.

@sbaldu from CERN and a maintainer of the header only performance portable C++ library  [alpaka](https://github.com/alpaka-group/alpaka) , is working on this, and will be working on other tasks for `xsf` as well. `alpaka` is in a similar problem-space as `xsf` and there is potential for the libraries to be used fruitfully together.

lucascolley:
sounds good, welcome aboard @sbaldu 👋!

PR #283, opened 2026-09-21
MAINT: refactor math function wrappers into `xsf::cxx`

sbaldu:
This PR refactors the math function wrappers in `config.h`, moving them to the `xsf::cxx` namespace rather than injecting them inside `std`.

Closes issue #282.


lucascolley:
~let's squash merge this~ and add to https://github.com/scipy/xsf/blob/main/.git-blame-ignore-revs in an immediate follow-up

include/xsf/config.h
lucascolley, line 1: actually, can we separate the changes to this file out into a separate commit, and make sure to _not_ squash merge

steppi: That's right. There should be a separate commit for the `config.h` change, and `git-blame-ignore-revs` for the other commit that changes all of the other files. Thanks Lucas

sbaldu: Done, I separated the commit refactoring config.h and the one updating the function calls

sbaldu: Should I add the commit to `git-blame-ignore` here or in a separate PR?

lucascolley: a separate PR is fine

j-bowhay:
Obviously not within the scope of this PR but it would be cool to have a linter check for this (and maybe other xsf patterns)

steppi:
> Obviously not within the scope of this PR but it would be cool to have a linter check for this (and maybe other xsf patterns)

include/xsf/agm.h
```diff
@@ -27,23 +27,23 @@ namespace detail {
 
 // Arithmetic-geometric mean.
 XSF_HOST_DEVICE inline double agm(double a, double b) {
-    // sqrthalfmax is std::sqrt(std::numeric_limits<double>::max() / 2)
+    // sqrthalfmax is std::sqrt(cxx::numeric_limits<double>::max() / 2)
```
steppi: It seems like the intention was not to mix  `std` and `cxx` in this comment here. We should pick one. Either is fine in this context

steppi (commented):
Thanks @sbaldu. This is looking good. I've confirmed everything is working fine in CuPy. I don't see any serious blockers. 

A few minor things:

* There's one comment in `"agm.h"` that mixes `std` and `cxx` that should pick one.
* In `"xsf/include/xsf/gamma.h"` there's a function `gamma_ratio` that is missing `XSF_HOST_DEVICE` and uses `std::tgamma` that isn't in `config.h` yet. Let's add `XSF_HOST_DEVICE` to that one and add `tgamma` to `xsf::cxx` in `config.h`. I've checked that `tgamma` should be available.
* We should add more than just the double overloads for math functions in `config.h` now that everything goes through the wrappers. There are some places where native float32 arithmetic is used, though most of the library handles float32 by casting to double and back. In the future, I'd like to be able to support native float32 computation for as much as possible though.

Some future follow-up work that you don't have to worry about at least for now. I'd rather you work on the `relaxed_layout` for `mdspan`.

I found the following headers which still have direct use of `std` in them.

* "numbers.h"
* "hyperu.h"
* "log_exp.h"
* "log.h"
* "mathieu_legacy.h"
* "multivariate_normal.h"
* "legendre.h"
* "dual.h"
* "sph_bessel.h"
* "specfun.h"
* "gamma.h"
* "hyp0f1.h"
* "amos/amos.h"
* "sphd_wave.h"
* "airy.h"
* "bessel.h"
* "recur.h"
* "kelvin.h"
* "fresnel.h"
* "exp.h"
* "par_cyl.h"
* "specfun/specfun.h"
* "numpy.h"

most of these are CPU only, so should be moved to the `cpu` namespace/filepath alongside `"xsf/include/xsf/cpu/stats.h"`. One exception  is `"gamma.h"` which I already discussed above. Some of these like `"numbers.h"` could be made to work on GPU pretty easily I think.

After splitting  the above headers off as cpu-only, we should add lint checks as suggested by @j-bowhay . These could ensure that at least:

1) XSF_HOST_DEVICE is present on all functions and function templates, except those in the `xsf::cpu` namespace
2) Headers other than `config.h` and those for things in the CPU namespace should only have internal includes from within `xsf`.
3) `std` is not used directly in headers other than `config.h` and those for things in the cpu namespace.

include/xsf/gamma.h
steppi: To stay in this file, this needs `XSF_HOST_DEVICE` and for `tgamma` to be added to `xsf::cxx` in `config.h`

include/xsf/config.h
```diff
@@ -64,95 +81,535 @@
...
-XSF_HOST_DEVICE inline double ldexp(double num, int exp) { return ::ldexp(num, exp); }
-XSF_HOST_DEVICE inline double fmod(double x, double y) { return ::fmod(x, y); }
-XSF_HOST_DEVICE inline double nextafter(double from, double to) { return ::nextafter(from, to); }
+template <typename T>
+using numeric_limits = std::numeric_limits<T>;
 #endif
 
-template <typename T, size_t N>
-using array = cuda::std::array<T, N>;
+XSF_HOST_DEVICE inline double abs(double num) {
```
steppi: Now that we go through the wrappers even on CPU, let's add at least `float` overloads as well for all of these math functions.

steppi: Let's do this in a follow up actually

include/xsf/gamma.h
steppi: Can be done in a follow up I think

include/xsf/agm.h
sbaldu: Thanks, I fixed the comment

steppi:
I'm going to go ahead and merge this. Thanks @sbaldu! This is going to create merge conflicts in pretty much all outstanding PRs, but hopefully it will not cause too much trouble to deal with them.

